PR #1, opened 2020-10-22
Enable Shell based unit tests

spbrogan:


Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -1,10 +1,37 @@
...
 #
 # Copyright (c) Microsoft Corporation
-#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
 
+#####
+# env Variables used in this plugin
+#
+# STARTUP_NSH_DIRTY:
```
out0xb2, line 12: Misleading name.  To match current code behavior, should be renamed DO_NOT_FORMAT_VIRTUAL_DRIVE, because it has nothing to do with startup.nsh.  The current code wipes the virtual drive even when **MAKE**_STARTUP_NSH is not TRUE.  I would prefer that, by default, if no parameters are specified, that the virtual drive is **NOT** formatted.  If the user specifies to make a startup.nsh, okay, then the default behavior can include wiping the drive unless this flag is also set to TRUE ...

spbrogan:
this was resolved in another PR

PR #7, opened 2021-01-04
add readme change to test for bug

spbrogan:


spbrogan:
closing.  the path for the DSC says it must be workspace relative....that is not really true...As long as the Pr_eval settings manager passing packages path (like it should) then the path can be relative.  Q35 does this correctly.

PR #89, opened 2022-10-14
[Rebase & FF] Integrate MM_SUPV into Q35 platform

kuqin12:
This change updated original OVMF package code to extend the support for MM_STANDALONE drivers for x86 systems. The dynamic PCDs are replaced by check-and-assert logic.

The system is verified bootable to Windows on Q35 system.


Platforms/QemuQ35Pkg/PlatformPei/Platform.c
```diff
@@ -433,18 +450,22 @@ MiscInitialization (
...
-                    PcdOvmfHostBridgePciDevId,
-                    MICROVM_PSEUDO_DEVICE_ID
-                    );
+      PcdStatus = EFI_UNSUPPORTED;
+      // MU_CHANGE: Remove dynamic PCD set to support usage in Standalone MM
+      // PcdStatus = PcdSet16S (
+      //               PcdOvmfHostBridgePciDevId,
+      //               MICROVM_PSEUDO_DEVICE_ID
+      //               );
       ASSERT_RETURN_ERROR (PcdStatus);
```
os-d, line 459: Maybe we never hit these cases, but won't this and the next one assert 100% of the time?

Platforms/QemuQ35Pkg/QemuFlashFvbServicesRuntimeDxe/FwBlockService.c
```diff
@@ -1129,7 +1131,10 @@ FvbInitialize (
   //
   InstallVirtualAddressChangeHandler ();
 
-  PcdStatus = PcdSetBoolS (PcdOvmfFlashVariablesEnable, TRUE);
-  ASSERT_RETURN_ERROR (PcdStatus);
+  // MU_CHANGE: Abstract dynamic PCD set to support Standalone MM
+  // PcdStatus = PcdSetBoolS (PcdOvmfFlashVariablesEnable, TRUE);
+  // ASSERT_RETURN_ERROR (PcdStatus);
```
os-d, line 1136: For all the commented out code segments can we just remove that code? Or do you think there is value in keeping the old code as reference for what was changed?

Platforms/QemuQ35Pkg/PlatformPei/Platform.c
kuqin12: Yes :) It is basically saying that we are not supporting types other than Q35.

Platforms/QemuQ35Pkg/QemuFlashFvbServicesRuntimeDxe/FwBlockService.c
kuqin12: I think it makes sense now. This was made before the removal of OvmfPkg, I will go through them and see if what can be cleaned up in that regard.

apop5 (approved):

Platforms/QemuQ35Pkg/QemuFlashFvbServicesRuntimeDxe/FwBlockService.h
```diff
@@ -194,4 +194,9 @@ SetPcdFlashNvStorageBaseAddresses (
   VOID
   );
 
+VOID
+UpdateOvmfFlashVariablesEnable (
```
kenlautner, line 198: Any particular reason why you have Ovmf in the name of this new function?  Given that we've gotten rid of the package.

Platforms/QemuQ35Pkg/QemuQ35Pkg.ci.yaml
```diff
@@ -42,7 +43,7 @@
             "SetupDataPkg/SetupDataPkg.dec",
             "MfciPkg/MfciPkg.dec",
             "OemPkg/OemPkg.dec",
-            "QemuQ35Pkg/QemuQ35Pkg.dec"
+            "QemuQ35Pkg/QemuQ35Pkg.dec",
```
kenlautner, line 46: This comma seems unnecessary.

Platforms/QemuQ35Pkg/QemuFlashFvbServicesRuntimeDxe/FwBlockService.h
kuqin12: Not anymore~ will update.

kenlautner (approved):

Platforms/QemuQ35Pkg/QemuQ35Pkg.ci.yaml
kuqin12: Removed.

Platforms/QemuQ35Pkg/QemuFlashFvbServicesRuntimeDxe/FwBlockService.h
kuqin12: Renamed to QEMU flash variable.

Platforms/QemuQ35Pkg/QemuFlashFvbServicesRuntimeDxe/FwBlockService.c
kuqin12: Went through all commits another round. The comments should be removed now.

PR #102, opened 2022-10-21
Remove OvmfPkg From CI Build

os-d:
With OvmfPkg removed from the repo, local CI builds fail if `-p QemuQ35Pkg` is not passed in because OvmfPkg is still referenced. The pipeline builds pass this parameter, which is why they do not fail.

Use the brackets because without them Python tries to iterate over the string and build a Q pkg, an e pkg, an m pkg, etc.


.pytool/CISettings.py
```diff
@@ -41,10 +41,7 @@ def GetPackagesSupported(self):
         ''' return iterable of edk2 packages supported by this build.
         These should be edk2 workspace relative paths '''
 
-        return (
-                "OvmfPkg",
-                "QemuQ35Pkg"
-                )
+        return {"QemuQ35Pkg"}
```
Javagedes, line 44:
```suggestion
        return ("QemuQ35Pkg",)
```

Javagedes (changes requested):

kuqin12 (approved):

.pytool/CISettings.py
os-d: Changed.

makubacki (approved):

Javagedes (approved):

apop5 (approved):

apop5 (approved):

apop5 (approved):

cfernald (approved):

PR #136, opened 2022-11-12
Implement SetPackages API so that invalid packages will raise exceptions rather than silently fail

spbrogan:
_Moving back to draft.   I want to see if the parameter validation can be handled by common logic in the invokable.  Below is a unfinished draft write up_

Update QemuQ35Pkg's platformbuild.py to implement the SetPackages API of the multipackage invocable Settings Manager to catch command line invokers that provide the wrong parameter and thus cause the stuart_pr_eval tool to execute incorrectly.

Stuart_PR_Eval is a multipackage aware tool which means that it has a cmd line parameter for the package to operate on. 
 For PR_Eval SetPackages has a default API that allows a subclass to not implement the function.


makubacki (approved):

makubacki:
If you're planning to squash, I recommend updating the PR description to explain the separate changes. If rebase & ff (which seems more appropriate), then the merge commit will need cleaned up and it would be nice for each to commit to have a description.

kuqin12 (approved):

PR #234, opened 2023-01-12
Add Test App Execution to Platform CI for VS Builds

TaylorBeebe:
## Description

To increase confidence in submodule updates, this PR adds the requirement that test apps included in the Q35 DSC successfully run during CI.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested in azure pipelines.

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuQ35Pkg/PlatformBuild.py
+++ b/Platforms/QemuQ35Pkg/PlatformBuild.py
@@ -32,6 +32,7 @@ failure_exempt_tests["MorLockFunctionalTestApp.efi"] = datetime.datetime(2022, 2
 failure_exempt_tests["MsWheaEarlyUnitTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
 failure_exempt_tests["VariablePolicyFuncTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
 failure_exempt_tests["DeviceIdTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["DxePagingAuditTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
 
 # Allow failure exempt tests to be ignored for 90 days
 FAILURE_EXEMPT_OMISSION_LENGTH = 90*24*60*60
@@ -442,10 +443,11 @@ class UnitTestSupport(object):
                         caseresult = "\t\t\tPASS"
                         level = logging.INFO
                         for result in case:
-                            if result.tag == 'failure' and not ignore_failure:
-                                failure_count += 1
+                            if result.tag == 'failure':
                                 level = logging.ERROR
                                 caseresult = "\t\tFAIL" + " - " + unescape(result.attrib['message'])
+                                if not ignore_failure:
+                                    failure_count += 1
                         logging.log( level, caseresult)
             except Exception as ex:
                 logging.error("Exception trying to read xml." + str(ex))
--- a/Platforms/QemuQ35Pkg/PlatformBuild.py
+++ b/Platforms/QemuQ35Pkg/PlatformBuild.py
@@ -9,7 +9,7 @@ import logging
 import io
 import shutil
 import glob
-import time
+import datetime
 import xml.etree.ElementTree
 import tempfile
 import uuid
@@ -22,6 +22,16 @@ from edk2toolext.invocables.edk2_update import UpdateSettingsManager
 from edk2toolext.invocables.edk2_pr_eval import PrEvalSettingsManager
 from edk2toollib.utility_functions import RunCmd, GetHostInfo
 from typing import Tuple
+
+# Declare test whose failure will not return a non-zero exit code
+failure_exempt_tests = {}
+failure_exempt_tests["BaseCryptLibUnitTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["BootAuditTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["LineParserTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["MorLockFunctionalTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["MsWheaEarlyUnitTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["VariablePolicyFuncTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
+failure_exempt_tests["DeviceIdTestApp.efi"] = datetime.datetime(2022, 2, 1, 0, 0, 0)
 
     # ####################################################################################### #
     #                                Common Configuration                                     #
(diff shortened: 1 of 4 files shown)


Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
```diff
@@ -1461,12 +1461,12 @@ PlatformSmmProtectionsTestLib|UefiTestingPkg/Library/PlatformSmmProtectionsTestL
     UefiTestingPkg/AuditTests/PagingAudit/UEFI/DxePagingAuditTestApp.inf
     UefiTestingPkg/FunctionalSystemTests/ExceptionPersistenceTestApp/ExceptionPersistenceTestApp.inf
     # UefiTestingPkg/FunctionalSystemTests/VarPolicyUnitTestApp/VarPolicyUnitTestApp.inf
-    CryptoPkg/Test/UnitTest/Library/BaseCryptLib/BaseCryptLibUnitTestApp.inf {
-      <PcdsPatchableInModule>
-        #Turn off Halt on Assert and Print Assert so that libraries can
-        #be tested in more of a release mode environment
-        gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x0E
-    }
+    # CryptoPkg/Test/UnitTest/Library/BaseCryptLib/BaseCryptLibUnitTestApp.inf {
```
spbrogan, line 1464: as discussed, let leave this test enabled.  
In the processing of test results, we should enable a path to ignore a result for known failures.  
Known failures should be a well defined table/list in python that includes a link to a bug.  
Maybe they should have a "last review date" so that if they are older than 3 months they have to be renewed.  

I am just making up stuff here but you get the idea.

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -305,27 +322,67 @@ def FlashRomImage(self):
         startup_nsh = StartUpScriptManager()
         run_tests = (self.env.GetValue("RUN_TESTS", "FALSE").upper() == "TRUE")
         output_base = self.env.GetValue("BUILD_OUTPUT_BASE")
+        nshpath = os.path.join(output_base, "startup.nsh")
         shutdown_after_run = (self.env.GetValue("SHUTDOWN_AFTER_RUN", "FALSE").upper() == "TRUE")
         empty_drive = (self.env.GetValue("EMPTY_DRIVE", "FALSE").upper() == "TRUE")
+        drive_path = (self.env.GetValue("DRIVE_PATH", None))
+        drive_is_shared_dir = self.env.GetValue("USE_SHARED_DIR", None)
+        if drive_is_shared_dir is not None:
+            drive_is_shared_dir = (drive_is_shared_dir.upper() == "TRUE")
```
Javagedes, line 331:
```suggestion
        drive_is_shared_dir = (self.env.GetValue("USE_SHARED_DIR", "FALSE") == "TRUE")
```

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -305,27 +322,67 @@ def FlashRomImage(self):
...
+            elif os.path.basename(drive_path).endswith(".vhd"):
+                if drive_is_shared_dir != True:
+                    drive_is_shared_dir = False
+                    virtual_drive = VirtualDriveManager(drive_path, self.env)
+                else:
+                    raise Exception("DRIVE_PATH is a VHD file, but USE_SHARED_DIR is set to TRUE")
+            else:
+                raise Exception("DRIVE_PATH is not a directory or a VHD file")
+        elif drive_is_shared_dir is None:
+            drive_is_shared_dir = False
```
Javagedes, line 353: Not necessary if you follow my suggested change above.

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -305,27 +322,67 @@ def FlashRomImage(self):
...
+        test_regex = self.env.GetValue("TEST_REGEX", "")
+
+        if drive_path is not None: 
+            if not os.path.exists(drive_path):
+                raise Exception("DRIVE_PATH does not exist")
+            elif os.path.isdir(drive_path):
+                if drive_is_shared_dir != False:
+                    drive_is_shared_dir = True
+                elif drive_is_shared_dir is False:
+                    raise Exception("DRIVE_PATH is a directory, but USE_SHARED_DIR is set to FALSE")
```
Javagedes, line 343:
```suggestion
            elif os.path.isdir(drive_path) and drive_is_shared_dir is False:
                raise Exception("DRIVE_PATH is a directory, but USE_SHARED_DIR is set to FALSE")
```

Platforms/QemuQ35Pkg/PlatformBuild.py
Javagedes, line 353: This is messy, I'm suggesting a cleanup. I believe it hits what you're trying to do, but you'll need to verify, if you are willing to take on the suggestion.

```suggestion
        ut = UnitTestSupport(os.path.join(output_base, "X64"))
        test_regex = self.env.GetValue("TEST_REGEX", "")
        
        drive_path = (self.env.GetValue("DRIVE_PATH", None))
        drive_is_shared_dir = (self.env.GetValue("USE_SHARED_DIR", "FALSE") == "TRUE")
        virtual_drive = None

        if drive_path is not None: 
            if not os.path.exists(drive_path):
                raise Exception("DRIVE_PATH does not exist")
            elif os.path.isdir(drive_path) and drive_is_shared_dir is False:
                raise Exception("DRIVE_PATH is a directory, but USE_SHARED_DIR is set to FALSE")
            elif os.path.basename(drive_path).endswith(".vhd"):
                if drive_is_shared_dir is True:
                    raise Exception("DRIVE_PATH is a VHD file, but USE_SHARED_DIR is set to TRUE")
                virtual_drive = VirtualDriveManager(drive_path, self.env)                    
            else:
                raise Exception("DRIVE_PATH is not a directory or a VHD file")
```

Javagedes: You could also be much more simple, but slightly more unsafe by doing what you did in QemuRunner.py

TaylorBeebe: Shareddir doesn't work well in the pipeline with many reads/writes so I'm removing the last several commits which added shareddir for testing

kuqin12 (approved):

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -393,8 +416,15 @@ def report_results(self, virtualdrive) -> int:
...
         #now parse the xml for errors
         failure_count = 0
-        logging.info("UnitTest Completed")
+        logging.info("UnitTest(s) Completed")
         for unit_test in self.test_list:
+            ignore_failure = False
+            if (os.path.basename(unit_test) in failure_exempt_tests.keys()):
+                now = datetime.datetime.now()
+                last_ignore_time = failure_exempt_tests[os.path.basename(unit_test)]
+                if (now - last_ignore_time).total_seconds() > FAILURE_EXEMPT_OMISSION_LENGTH:
```
spbrogan, line 425: not sure i understand this.  Doesn't this read
if the difference between now and "last ignore time" is greater than OMISSION_LENGTH then ignore it?  
This seems like it would cause a failure to be ignored after 90 days. Don't we want the inverse?

apop5 (approved):

PR #445, opened 2023-03-25
[REBASE && FF] Add Support for Virtual Drive Testing on Linux

TaylorBeebe:
Add virtual drive support for linux distros and run unit testing during CI for Linux GCC5 Q35 and SBSA, and Windows Q35.

### DIFF
--- a/Platforms/QemuQ35Pkg/PlatformBuild.py
+++ b/Platforms/QemuQ35Pkg/PlatformBuild.py
@@ -13,6 +13,7 @@ import datetime
 import xml.etree.ElementTree
 import tempfile
 import uuid
+import string
 
 from edk2toolext.environment import shell_environment
 from edk2toolext.environment.uefi_build import UefiBuilder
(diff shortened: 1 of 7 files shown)


Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -25,14 +25,14 @@
 
 # Declare test whose failure will not return a non-zero exit code
 failure_exempt_tests = {}
-failure_exempt_tests["BaseCryptLibUnitTestApp.efi"] = datetime.datetime(2023, 3, 7, 0, 0, 0)
```
os-d, line 28: Is this being removed because it fails? If so, is there a line of sight to getting it fixed?

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -461,24 +444,24 @@ def report_results(self, virtualdrive) -> int:
 class VirtualDriveManager(object):
```
os-d, line 444: Worth updating name to `WindowsVirtualDriveManager` to distinguish it from the Linux one (and not that it is a base implementation)?

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -314,68 +314,51 @@ def FlashRomImage(self):
...
+        if test_regex != "":
+            ut.set_test_regex(test_regex)
+            ut.find_tests()
+            ut.copy_tests_to_virtual_drive(VirtualDrive)
 
-            if os.path.exists(VirtualDrivePath) and empty_drive:
-                shutil.rmtree(VirtualDrivePath)
+        if run_tests:
+            if test_regex == "":
+                logging.warning("No tests specified using TEST_REGEX flag but RUN_TESTS is TRUE")
```
Javagedes, line 342: I think default behavior would be that TEST_REGEX is set to run all tests if RUN_TESTS is true. It seems silly that someone that wants to run all tests would need to do RUN_TESTS=TRUE TEST_REGEX=**/* (or whatever the proper regex is)

If that is the default behavior by the regex being "", then I don't think you need to throw a warning at the user. This would be expected behavior in my opinion

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -493,9 +476,55 @@ def GetFileContent(self, VirtualFilePath, HostFilePath: os.PathLike=None):
...
+        
+        # Create an mtools config file to virtually map the image to a drive letter
+        RunCmd("echo", "mtools_skip_check=1 > ~/.mtoolsrc")
+        RunCmd("echo", f"drive {self.drive_letter}: >> ~/.mtoolsrc")
+        RunCmd("echo", f"\"  file=\\\"{self.drive_path}\\\" exclusive\" >> ~/.mtoolsrc")
+
+        return 0
+
+    def AddFile(self, HostFilePath:os.PathLike):
+        # file_name = os.path.basename(HostFilePath)
```
os-d, line 509: comment unneeded?

Platforms/QemuQ35Pkg/PlatformBuild.py
Javagedes: If that is the default behavior by the regex being "", then I don't think you need to throw a warning at the user. This would be expected behavior in my opinion

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -493,9 +476,55 @@ def GetFileContent(self, VirtualFilePath, HostFilePath: os.PathLike=None):
             return f.read()
 
     def ExtractFile(self, VirtualFilePath, HostFilePath:os.PathLike):
-        ret = RunCmd("FileExtract", f"{VirtualFilePath} {HostFilePath} {self.path_to_vhd}")
+        ret = RunCmd("FileExtract", f"{VirtualFilePath} {HostFilePath} {self.drive_path}")
+        return ret
+
+class LinuxVirtualDriveManager(object):
+
+    def __init__(self, img_path:os.PathLike,  drive_letter:str="x"):
```
os-d, line 484: I understand we want a default for the CI build, but is this potentially dangerous if a user has a drive mounted at X: already and didn't change the default? I.e. is it going to overwrite the drive? If so, perhaps some safety around this would be nice (or just a check if X: already exists and failing).

Platforms/QemuQ35Pkg/PlatformBuild.py
Javagedes, line 1: This testing logic is copy+paste in two places and possibly more if we develop more platforms. I would make this a Helper Plugin in QemuPkg. (Similar to [QemuRunner ](https://github.com/microsoft/mu_tiano_platforms/tree/main/Platforms/QemuQ35Pkg/Plugins/QemuRunner)).


I also think at this point the QemuRunner plugin should be moved to QemuPkg. I can do that in a PR if needed

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
```diff
@@ -1264,84 +1264,70 @@
...
+  # UefiTestingPkg/AuditTests/TpmEventLogAudit/TpmEventLogAuditTestApp.inf # DOESN'T PRODUCE OUTPUT
+  # UefiTestingPkg/AuditTests/UefiVarLockAudit/UEFI/UefiVarLockAuditTestApp.inf # DOESN'T PRODUCE OUTPUT
+  UefiTestingPkg/FunctionalSystemTests/ExceptionPersistenceTestApp/ExceptionPersistenceTestApp.inf
+  UefiTestingPkg/FunctionalSystemTests/MemmapAndMatTestApp/MemmapAndMatTestApp.inf
+  UefiTestingPkg/FunctionalSystemTests/MorLockTestApp/MorLockTestApp.inf
+  # UefiTestingPkg/FunctionalSystemTests/SmmPagingProtectionsTest/App/SmmPagingProtectionsTestApp.inf # NOT APPLICABLE TO Q35
+  UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/App/MemoryProtectionTestApp.inf
+  UefiTestingPkg/FunctionalSystemTests/SmmPagingProtectionsTest/Smm/SmmPagingProtectionsTestSmm.inf
+  UefiTestingPkg/FunctionalSystemTests/MemoryProtectionTest/Smm/MemoryProtectionTestSmm.inf
+  # UefiTestingPkg/AuditTests/PagingAudit/UEFI/DxePagingAuditDriver.inf # TEST RUN VIA APPLICATION
```
os-d, line 1322: The test is run via the application, but the driver is still needed, right? Why are we commenting it out? To wait for the change to the mu_plus test? In which case, can't we leave this driver in and just switch out the app later?

Platforms/QemuSbsaPkg/PlatformBuild.py
```diff
@@ -499,6 +515,51 @@ def ExtractFile(self, VirtualFilePath, HostFilePath:os.PathLike):
         ret = RunCmd("FileExtract", f"{VirtualFilePath} {HostFilePath} {self.path_to_vhd}")
         return ret
 
+class LinuxVirtualDriveManager(object):
```
os-d, line 518: Would this be useful to have at a common python level (say in QemuPkg) so that we don't maintain multiple copies? Probably a lot of the python could be shared, so perhaps a separate PR could deal with all of it.

os-d (approved):

Platforms/QemuQ35Pkg/PlatformBuild.py
TaylorBeebe: I'm not sure I understand.

TEST_REGEX defines what tests are copied into the virtual drive and RUN_TESTS causes the tests to be added to the startup.nsh script for automated execution. There are cases where I only want to copy tests over and not have them run automatically but never a case where I want to run tests automatically but didn't actually copy any over. There's also no way to define what a test is other than by filename and module type (UEFI_APPLICATION).

Platforms/QemuQ35Pkg/PlatformBuild.py
TaylorBeebe: Ken fixed it a few weeks ago

Platforms/QemuQ35Pkg/PlatformBuild.py
TaylorBeebe: Yes, will remove

Platforms/QemuQ35Pkg/PlatformBuild.py
os-d: Oh, I see, my misunderstanding. Not each test goes here, just ones that are currently allowed to fail.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
TaylorBeebe: The driver isn't at all needed to run the paging audit through the shell. The driver is only necessary if the user wants to trigger the audit on the installation of some protocol.

Platforms/QemuQ35Pkg/PlatformBuild.py
Javagedes: Understood. I didn't think about the fact that you may want to bring tests over but not run them. I also thought had a way of knowing what tests we were compiling, and the test regex was just filtering on that test list.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
os-d: Ah, I see, I did not know that. Maybe worth removing this or expanding the comment?

Platforms/QemuQ35Pkg/PlatformBuild.py
TaylorBeebe: Yes, the only real difference is the folder being checked. Do you have bandwidth to do the unification? I can if not. Either way, I'd prefer to stay relatively close to the original intent of this effort (to bring virtual drive support to Linux and light up unit tests in CI) for this PR and simplify the structure of the platform scripts the next effort.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
TaylorBeebe: I put a powershell script to discover the test apps in each of the platform DSC files. For someone who updates the test apps later, they should be able to run the script and compare it to what's already in the DSC to identify NEW test apps. For this reason, I don't want to remove this from the test list.

Also, I think that the existing comment is sufficient to say why it's not necessary.

os-d: Okay, I'm fine with it as is. The existing comment makes me think that this is a deprecated way of running the test, i.e. now the test is run via application, so it feels unclear why the commented out driver is there.

Platforms/QemuQ35Pkg/PlatformBuild.py
Javagedes: Yes, I have the bandwidth. I can do it after this PR is merged.

Platforms/QemuQ35Pkg/PlatformBuild.py
TaylorBeebe: Updated

Platforms/QemuQ35Pkg/PlatformBuild.py
TaylorBeebe: Added a function find_unused_drive_letter() which will query the mtab file for mounted drives.

PR #475, opened 2023-04-10
Transition virtual drive functionality to a Helper Plugin

Javagedes:
## Description

With the creation of QemuSbsaPkg, nearly all of the Virtual hard drive functionality (hard drive management, test management, startup script management) had to be duplicated between QemuSbsaPkg and QemuQ35Pkg. This commit moves all of this functionality into a plugin found under QemuPkg/Plugins/VirtualDriveManager for maintainabilities sake.

Additionally provides some QOL improvements:

1. Error handling RunCmd's use for managing virtual drives
2. Displaying debug information to the user on CLI settings
3. Checking if a test has been run so the image does not need to be loaded and executed before being skipped. (An issue for tests that reset)
4. Updated the way we ignore exempt tests to make it more comprehensible.
5. Used RunCmd rather than popen for finding the unused drive on a linux system
6. Check the result of grep rather than the output of grep as it will return 0/1/2 for files_found/no_files_found/error occurred (respectively).
7. Set the virtual drive path in SetPlatformEnvAfterTarget rather then in FlashRom

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified testing continued to work on Windows/Linux QemuQ35Pkg and Linux QemuSbsaPkg

## Integration Instructions

N/A


TaylorBeebe:
Can reset_tests and the failure exempt tests for each platform be moved to a common file for when they need to be edited again?

Javagedes:
> Can reset_tests and the failure exempt tests for each platform be moved to a common file for when they need to be edited again?

@TaylorBeebe They can be, but I left them in their respective files because the lists were different between Q35 and SBSA

TaylorBeebe:
> > Can reset_tests and the failure exempt tests for each platform be moved to a common file for when they need to be edited again?
> 
> @TaylorBeebe They can be, but I left them in their respective files because the lists were different between Q35 and SBSA

The reset tests were moved to the plugin so now there are two places that may need to be changed as more tests are added. Can one or two functions be added to the helper plugin which return the relevant failure omission list for the platform?

Javagedes:
> > > Can reset_tests and the failure exempt tests for each platform be moved to a common file for when they need to be edited again?
> > 
> > 
> > @TaylorBeebe They can be, but I left them in their respective files because the lists were different between Q35 and SBSA
> 
> The reset tests were moved to the plugin so now there are two places that may need to be changed as more tests are added. Can one or two functions be added to the helper plugin which return the relevant failure omission list for the platform?

I removed the need for reset_tests by adding if statements around each test that checks for the existence of the appropriate JUNIT file. Now the only thing that needs to be updated is the exempt tests, which are in their respective platform build files. Let me know if you are okay with that :)

Javagedes:
The startup script is formatted as such:

image

.github/workflows/Build-UefiExt.yaml
cfernald, line 1: The only changes to this is formatting right? Perhaps this file should be dropped from the PR for clarity.

TaylorBeebe:
> > > > Can reset_tests and the failure exempt tests for each platform be moved to a common file for when they need to be edited again?
> > > 
> > > 
> > > @TaylorBeebe They can be, but I left them in their respective files because the lists were different between Q35 and SBSA
> > 
> > 
> > The reset tests were moved to the plugin so now there are two places that may need to be changed as more tests are added. Can one or two functions be added to the helper plugin which return the relevant failure omission list for the platform?
> 
> I removed the need for reset_tests by adding if statements around each test that checks for the existence of the appropriate JUNIT file. Now the only thing that needs to be updated is the exempt tests, which are in their respective platform build files. Let me know if you are okay with that :)

This work in most cases, but someone will end up being confused at some point when their test isn't running because "EMPTY_DRIVE=TRUE" wasn't in their build command. Is there a way to mitigate this?

TaylorBeebe:
> > > > > > Can reset_tests and the failure exempt tests for each platform be moved to a common file for when they need to be edited again?
> > > > > 
> > > > > 
> > > > > @TaylorBeebe They can be, but I left them in their respective files because the lists were different between Q35 and SBSA
> > > > 
> > > > 
> > > > The reset tests were moved to the plugin so now there are two places that may need to be changed as more tests are added. Can one or two functions be added to the helper plugin which return the relevant failure omission list for the platform?
> > > 
> > > 
> > > I removed the need for reset_tests by adding if statements around each test that checks for the existence of the appropriate JUNIT file. Now the only thing that needs to be updated is the exempt tests, which are in their respective platform build files. Let me know if you are okay with that :)
> > 
> > 
> > This work in most cases, but someone will end up being confused at some point when their test isn't running because "EMPTY_DRIVE=TRUE" wasn't in their build command. Is there a way to mitigate this?
> 
> Yes, I could add a file another script that deletes all xml files. The startup.nsh would call this file on the initial startup, then delete its self so that for any additional boots due to a reset, it would not run, leaving the current xml results safe and sound.
> 
> HOWEVER, I think we are protecting the user to much in this scenario. I think the preferred course of action is:
> 
> 1. EMPTY_DRIVE=FALSE as default (already happening)
> 2. If the user specifies RUN_TESTS=TRUE, we will automatically set EMPTY_DRIVE=TRUE.
> 3. If the user really does not want to empty their drive when running tests, thenthey can specify EMPTY_DRIVE=FALSE on the CLI, which will take precedence. In that situation, we already throw a warning at the user specifying: `EMPTY_DRIVE=FALSE. Old files can persist, could effect test results.`. I could make that warning a little better and say that any tests that have already run will not run again.

This is probably purely a preference issue, but I prefer to maximize for behavior consistency to reduce user frustration and the possibility of wasting time due to an unknown nuance. I think we can keep your update to check for the XML existence by renaming each XML after running all the tests.

I think it would just be a few lines to update. Add the following to the write_out function:
```
for l in self._lines:
  nsh.write(f"mv {l[:-4]}_JUNIT.XML {l[:-4]}_JUNIT_RESULT.XML\n")
```

And in report_results update the line to:
```
xml_result_file = os.path.basename(unit_test)[:-4] + "_JUNIT_RESULT.XML"
```

This should allow the XML check to still work without needing to force EMPTY_DRIVE=TRUE when RUN_TESTS=TRUE. Let me know what you think.

TaylorBeebe:
> 

Ah, yes I was looking at the original scripts when suggesting the change. Yours makes more sense - thanks! 😊

TaylorBeebe:
> > > 
> > 
> > 
> > Ah, yes I was looking at the original scripts when suggesting the change. Yours makes more sense - thanks! 😊
> 
> I had to also add in a line to delete the .dat files, or the tests will never run again because they use the .dat files to determine the status of the test. but the changes are implemented now :)

Because some test apps (like the paging audit) produce .dat files as artifacts, can you update the command to remove *_Cache.dat?

TaylorBeebe (approved):

cfernald (approved):

Javagedes:
> > > > 
> > > 
> > > 
> > > Ah, yes I was looking at the original scripts when suggesting the change. Yours makes more sense - thanks! 😊
> > 
> > 
> > I had to also add in a line to delete the .dat files, or the tests will never run again because they use the .dat files to determine the status of the test. but the changes are implemented now :)
> 
> Because some test apps (like the paging audit) produce .dat files as artifacts, can you update the command to remove *_Cache.dat?

Yes, not a problem, I've updated the PR!

kuqin12 (approved):

QemuPkg/Plugins/VirtualDriveManager/VirtualDriveManager.py
```diff
@@ -0,0 +1,410 @@
...
+            drive (VirtualDrive): The virtual drive to add the tests to.
+            auto_run (Boolean): Whether or not to run tests automatically.
+            auto_shutdown (Boolean): Whether or not to shutdown after tests have completed.
+        """
+        drive.add_files(test_list)
+
+        if auto_run:
+            tests = []
+            # Execute all tests
+            for test in test_list:
```
spbrogan, line 356: what is going on here?  This looks like some method to get around the "cache.dat" stuff.

QemuPkg/Plugins/VirtualDriveManager/VirtualDriveManager.py
```diff
@@ -0,0 +1,410 @@
...
+                tests.append(f"if not exist {test.stem}_JUNIT.XML then")
+                tests.append(f"    {test.name}")
+                tests.append("endif")
+            
+            # Tests have finished, no more restarts
+            # Remove any old test results and reset test status
+            tests.append("rm *_JUNIT_RESULT.XML")
```
spbrogan, line 363: and most importantly here

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -318,282 +327,74 @@ def __SetEsrtGuidVars(self, var_name, guid_str, desc_string):
...
+            virtual_drive.wipe()
+
+        if not virtual_drive.exists():
+            virtual_drive.make_drive()
+
+        # Add tests if requested, auto run if requested
+        if test_regex != "":
+            test_list = []
+            for pattern in test_regex.split(","):
+                test_list.extend(Path(output_base, "X64").glob(pattern))
```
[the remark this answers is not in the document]
Javagedes: Sure, I can glob from the output_base. This was just the previous logic.

Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -318,282 +327,74 @@ def __SetEsrtGuidVars(self, var_name, guid_str, desc_string):
...
             outstream = StringIO()
-            # See git-describe docs for a breakdown of this command output
-            ret = RunCmd("git", f'describe {commithash} --tags', outstream=outstream)
+            ret = RunCmd("git", f'describe {head["HexSha"]} --tags', outstream=outstream)
             if ret == 0:
                 version = outstream.getvalue().strip()
+        else:
+            version = "unknown"
 
         self.env.SetValue("VERSION", version, "Set Version value")
```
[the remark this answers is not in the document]
Javagedes: This was previous logic, I just changed it in this PR to use repo_details() to get the commit hash instead of RunCmd. I could move it to the qemu_runner plugin as it's the only consumer of the version.

QemuPkg/Plugins/VirtualDriveManager/VirtualDriveManager.py
spbrogan: ok.  after reading the PR history i now understand this.  Might make sense to add a comment to explain the complexity in the startup nsh and the solution to optimize the tests running

Javagedes: There are two parts to this:

1. This is a solution to reduce test times by preventing the efi image from loading if the test has already run. Before this change, any test that requires a restart would force the system to reload all previous test images and execute them. While the tests are skipped due to the presence of a _Cache.dat file, it still takes time to load the image.

1. We want to remove the *_Cache.dat file at the end of the run so that future runs (where EMPTY_DRIVE=FALSE) will actually run the test instead of loading the test, seeing that it has already run, and skipping.

PR #528, opened 2023-05-01
Output Results of Failure Exempt Tests on Q35 and SBSA

TaylorBeebe:
## Description

The update to use a helper plugin for virtual drive managment changed the way we filter tests. Instead of report_results() checking for exempt tests during the report, the Q35 and SBSA platforms would filter out exempt tests from the test list passed to report_results(). As a result, the exempt unit tests would not have their XMLs copied to unit_test_results or have a summary of their results dumped via logging even if the tests passed.

This PR creates a second list containing only the exempt tests and passes that to report_results(). By ignoring the output of report_results() when passing in the exempt test list, we effectively ignore the failures.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Running on Q35 and SBSA

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuQ35Pkg/PlatformBuild.py
+++ b/Platforms/QemuQ35Pkg/PlatformBuild.py
@@ -397,7 +397,9 @@ class PlatformBuilder(UefiBuilder, BuildSettingsManager):
 
         # Filter out tests that are exempt
         tests = list(filter(lambda file: file.name not in FET or not (now - FET.get(file.name)).total_seconds() < FEOL, test_list))
-        
+        tests_exempt = list(filter(lambda file: file.name in FET and (now - FET.get(file.name)).total_seconds() < FEOL, test_list))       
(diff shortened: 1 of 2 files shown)


Javagedes:
While this will not return failure count and cause stuart_build to fail, it will still log the failures as an error on the CLI. Is this your intention?

TaylorBeebe:
> While this will not return failure count and cause stuart_build to fail, it will still log the failures as an error on the CLI and log file. Is this your intention?

Yes, that is my intent. As long as FlashRomImage() doesn't return a non-zero result the "Run to Shell" step in the pipelines won't fail.

Javagedes (approved):

makubacki (approved):

PR #625, opened 2023-07-05
Use correct Openssl inf for SBSA

kenlautner:
## Description

Changed the openssl inf version used to be the full version to match expected functionality.    Additionally disabled NETWORK_ISCSI_ENABLE as we weren't using it.

Extra crypto functionality include:
1. Every SHA hash function.  The default openssl version support SHA256 only.
2. X509 full functionality (this is connected with item 3)
3. ECC for TLS and other crypto features (item #2)

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passed BaseCryptLib Unit Tests.

## Integration Instructions

N/A


makubacki:
@kenlautner, I know this might be obvious, but can you list the additional functionality brought in by the change to the PR description?

makubacki (approved):

TaylorBeebe (changes requested):
Is this change also supposed to resolve the BaseCryptLibUnitTest failure?

kenlautner:
> Is this change also supposed to res

kenlautner:
> Is this change also supposed to resolve the BaseCryptLibUnitTest failure?

Yes it is.  I see that it's still failing for Md5 Hashing which was removed from support.  We should probably discuss if we want to allow the use of Md5 hashing again or update the test.  It's legacy so we should probably add support in special cases such as this.

kuqin12 (approved):

kuqin12:
> > Is this change also supposed to resolve the BaseCryptLibUnitTest failure?
> 
> Yes it is. I see that it's still failing for Md5 Hashing which was removed from support. We should probably discuss if we want to allow the use of Md5 hashing again or update the test. It's legacy so we should probably add support in special cases such as this.

So Q35 is not failing this test? Is it because Q35 disabled the deprecated test, or because it is also failing but still in grace period?

kenlautner:
> > > Is this change also supposed to resolve the BaseCryptLibUnitTest failure?
> > 
> > 
> > Yes it is. I see that it's still failing for Md5 Hashing which was removed from support. We should probably discuss if we want to allow the use of Md5 hashing again or update the test. It's legacy so we should probably add support in special cases such as this.
> 
> So Q35 is not failing this test? Is it because Q35 disabled the deprecated test, or because it is also failing but still in grace period?

Q35 doesn't use MD5 and it is disabled by default in MU.  SBSA enables it by including NetworkBuildOptions.dsc.inc and NETWORK_ISCSI_ENABLE is TRUE.

TaylorBeebe (approved):

Issue #607, opened 2023-06-27
[Bug]: QEMU runner cannot boot to OS on qcow2 images


kuqin12:
### Is there an existing issue for this?

- [X] I have searched existing issues

### Current Behavior

After https://github.com/microsoft/mu_tiano_platforms/pull/489/files#diff-f29acc9b32cd66d4c8c4c23f631994df06954559b7de7978658a3a61ecf61e75, the QemuRunner.py will no longer take a qcow2 OS image to boot on QEMU.

For example, it used to take "PATH_TO_OS=<os_image_folder/os_image_name.qcow2>" to boot to the OS, now the image does not map to the file system protocols. Specifically, the `format=raw` is causing the file not being able to be mapped.

### Expected Behavior

Map the qcow2 image as before. Because the qcow2 images are more resilient to the sudden deaths of QEMU process and would not get corrupted in those cases.

### Steps To Reproduce

1. Build QemuQ35Pkg
2. Boot the built images using "PlatformBuild.py PATH_TO_OS=<os_image_name.qcow2> --flashonly", where the os_image_name.qcow2 can be made from vhdx images using `qemu-img convert -f vhdx -p -c -O qcow2 foo.vhdx foo.qcow2`
3. Notice that the system would attempt to boot to the qcow2 image and failed, then ended up in UEFI shell.

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): VS2022
- Targets Impacted: DEBUG, RELEASE
```

### Version Information

```text
https://github.com/microsoft/mu_tiano_platforms/pull/489/files#diff-f29acc9b32cd66d4c8c4c23f631994df06954559b7de7978658a3a61ecf61e75
```

### Urgency

Low

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

Need to know why only the vhd is chosen to be supported in the first place.

PR #635, opened 2023-07-14
QemuRunner: Add support for QCOW2 images

Flickdm:
# Preface

Please ensure you have read the [contribution docs](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md) prior
to submitting the pull request. In particular,
[pull request guidelines](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md#pull-request-best-practices).

## Description

This adds support for both VHD and QCOW2 images by using the extension to determine which storage rule should be used for qemu See [Issue 607](https://github.com/microsoft/mu_tiano_platforms/issues/607)

- [X] Impacts functionality?
  - Adds back functionality for qcow2 images
- [ ] Impacts security?
  - N/A
- [ ] Breaking change?
  - N/A
- [ ] Includes tests?
  - N/A
- [ ] Includes documentation?
  - N/A
  
## How This Was Tested
This was used to run both a vhd image and a cow2 image, to confirm backward compatibility the resultant string was also compared

## Integration Instructions
N/A

### DIFF
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -16,6 +16,7 @@ import subprocess
 import re
 import io
 import shutil
+from pathlib import Path
 from edk2toolext.environment import plugin_manager
 from edk2toolext.environment.plugintypes import uefi_helper_plugin
 from edk2toollib import utility_functions
@@ -91,10 +92,12 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
             # Potentially dealing with big daddy, give it more juice...
             args += " -m 8192"
 
+            file_extension = Path(path_to_os).suffix.lower().replace('"', '')
+            
             storage_rule = {
-                ".vhd": " -drive format=raw,index=0,media=disk,file=\"" + path_to_os + "\"",
-                ".qcow2": " -hda \"" + path_to_os + "\""
-            }.get(os.path.splitext(path_to_os)[1].replace('"', '').lower(), None)
+                ".vhd": f" -drive format=raw,index=0,media=disk,file=\"{path_to_os}\"",
+                ".qcow2": f" -hda \"{path_to_os}\""
+            }.get(file_extension, None)
 
             if storage_rule is None:
                 raise Exception("Unknown OS storage type: " + path_to_os)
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -90,8 +90,16 @@ def Runner(env):
         if path_to_os is not None:
             # Potentially dealing with big daddy, give it more juice...
             args += " -m 8192"
-            #args += " -hda \"" + path_to_os + "\""
-            args += " -drive format=raw,index=0,media=disk,file=\"" + path_to_os + "\""
+
+            storage_rule = {
```
makubacki, line 94: For new code additions, I suggest using f-strings so the substitution is less verbose and `pathlib` to get the extension via `.suffix` which is more straightforward to read.

Something like:

```python
file_extension = Path(path_to_os).suffix.lower().replace('"', '')

storage_rule = {
    ".vhd": f" -drive format=raw,index=0,media=disk,file=\"{path_to_os}\"",
    ".qcow2": f" -hda \"{path_to_os}\""
}.get(file_extension, None)
```

Makes it easier to read.

makubacki, line 94: For new code additions, I suggest using f-strings so the substitution is less verbose and `pathlib` to get the extension via `.suffix` which is more straightforward to read.

Something like:

```suggestion
            file_extension = Path(path_to_os).suffix.lower().replace('"', '')

            storage_rule = {
                ".vhd": f" -drive format=raw,index=0,media=disk,file=\"{path_to_os}\"",
                ".qcow2": f" -hda \"{path_to_os}\""
            }.get(file_extension, None)

```

Makes it easier to read.

Flickdm: Agreed, I only left it this way to keep compatibility with python 3.6 if that's not a concern I'm happy to make this change

makubacki: We're past that point now in a number of scripts.

Javagedes: pytools provides support for 3.9, 3.10, and 3.11 so any pytool scripts need only support those versions.

Javagedes (approved):

makubacki (approved):

PR #640, opened 2023-07-18
Introduce a standalone script for acquiring and running Project MU QEMU platforms.

cfernald:
## Description

Introduces the muemu.py script for running project MU virtual platforms outside of the context of the UEFI build environment.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested the script locally and part of private pipelines.

## Integration Instructions

N/A


apop5:
not stale.

Scripts/muemu.py
```diff
@@ -0,0 +1,257 @@
...
+    qemu_args += ["-drive",
+                  f"if=pflash,format=raw,unit=1,file={efi_fd},readonly=on"]
+
+    if args.cores != 4 and not args.force:
+        print("Arm64 image must use 4 cores! Setting to 4.")
+        args.cores = 4
+
+
+def run_qemu(qemu_args: List[str]):
+    if args.tpm:
```
makubacki, line 177: Set `swtpm_proc` to `None` here to prevent an "unbound error" on when the value is compared to `None` later in the function and `args.tpm` is `False`.

```suggestion
def run_qemu(qemu_args: List[str]):
    swtpm_proc = None

    if args.tpm:
```

Scripts/muemu.py
```diff
@@ -0,0 +1,267 @@
...
+                    help="Enables update to use the DEBUG firmware binaries.")
+parser.add_argument("--verbose", action="store_true",
+                    help="Enabled verbose script prints.")
+parser.add_argument("--force", action="store_true",
+                    help="Disables automatic correction of VM configurations.")
+parser.add_argument("--timeout", type=int, default=None,
+                    help="The number of seconds to wait before killing the QEMU process.")
+parser.add_argument("--qemuargs", type=str, default=None,
+                    help="Additional arguments to provide to QEMU, using the format \" -foo bar\". "
+                    "The preceding space is required for parsing reasons.")
```
cfernald, line 65: Looking for suggestions on this. Annoyingly argparse seems to really freak out when `-` is the first character of a string, and so i have to require a space at the beginning for it to read the whole string as one thing.

kuqin12: I think if you use `parse_known_args`, the result will become "known_args, left_over = parser.parse_known_args()"

PR #688, opened 2023-08-30
Add Rust support and DXE Rust driver to boot [Rebase & FF]

makubacki:
## Description

Includes changes to add Rust build support to the repo, include a Rust
DXE driver (`HelloWorldRustDxe`) in the `QemuQ35Pkg` and `QemuSbsaPkg`
build and boot, and associated pipeline changes.

Change summary:

---

**Add repo Rust infrastructure support**

Adds files to support:

- A cargo-make makefile
- Default Rust toolchain version
- Rust formatting configuration
- Rust host-based unit test CI execution

---

**Bump Common/MU from 2023020001.3.1 to 2023020001.4.0**

---

**Add HelloWorldRustDxe to QemuQ35Pkg and QemuSbsaPkg**

Adds the driver to both packages to demonstrate a Rust based UEFI
DXE driver executing during boot on a X64 and AARCH64 system.

A workspace Cargo.toml file is added for cargo make. Since this repo
is the "leaf of a dependency chain" in that it builds binaries that
are integrated into the platform build, the `Cargo.lock` file is
checked in per the guidance in:
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

---

**.azurepipelines: Add Rust support**

Updates the repo pipelines to support Rust builds.

---

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified CI and Platform builds locally and in CI
- Verified QemuQ35Pkg and QemuSbsaPkg builds
- Verified HelloWorldRustDxe dispatches during boot and prints
  the expected debug messages

## Integration Instructions

N/A

---

Note: The `MU_BASECORE` submodule will be updated after https://github.com/microsoft/mu_basecore/pull/545 completes.


os-d (approved):

Cargo.lock
Javagedes, line 1: I know in the rust community, that whether to include the Cargo.lock file or not is controversial. Just pointing this out in case it was an accident to include (I have no bias).

os-d: I wondered the same thing, binged it, accidentally clicked on the ChatGPT response, which told me for an end product binary, it should be included, but not for a library, etc. So I thought, ok, I guess it should be! :)

makubacki: I agree and I don't have a strong opinion. My understanding is that it makes sense for this repo since we would like to capture a dependency snapshot at the time binaries were built and tested in the platform build/boot.

If there is a reason not to check it in though, I'm happy to remove it as well.

makubacki: Sounds like, we'll go with keeping it for now.

makubacki:
Currently waiting on #689 so the mu_basecore submodule can be updated (the change needed for this PR in mu_basecore is after the UefiCpuLib breaking change).

apop5 (approved):

PR #718, opened 2023-09-28
Fixing multiple build instances usage on Linux environment

kuqin12:
## Description

Current script will try to update the mcopy configuration file. After QEMU run, the build script will then try to open the virtual disk indicated in this configuration. When there are multiple building instances on the same Linux system, we could end up with various conflicts, either opening up the incorrect image file, or opened up a file from other build jobs.

This change attempts to fix it by initiating a configuration file for each build for reference and update the environment variable to set up the configuration.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on both selfhost agents internal and here on the public pipeline builds.

## Integration Instructions

N/A.

### DIFF
--- a/QemuPkg/Plugins/VirtualDriveManager/VirtualDriveManager.py
+++ b/QemuPkg/Plugins/VirtualDriveManager/VirtualDriveManager.py
@@ -166,7 +166,7 @@ class LinuxVirtualDrive(VirtualDrive):
         
         # Create an mtools config file to virtually map the image to a drive letter
         RunCmd("echo", "mtools_skip_check=1 > ~/.mtoolsrc")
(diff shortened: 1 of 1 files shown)


os-d (approved):
Is there a similar issue for Windows that should be fixed?

kuqin12:
> Is there a similar issue for Windows that should be fixed?

Not for Windows. Windows build has been using tools from `mu-disc-utils` for img file manipulation, which is self-contained for each build session.

TaylorBeebe (approved):

PR #764, opened 2023-11-08
Bump Common/MU from 2023020003.1.0 to 2023020003.2.0

ProjectMuBot:
Bumps Common/MU from `2023020003.1.0` to `2023020003.2.0`

Introduces 2 new commits in [Common/MU](https://github.com/microsoft/mu_plus.git).

<details>
<summary>Commits</summary>
<ul>
<li><a href="https://github.com/microsoft/mu_plus/commit/20cb1703e1f9214dd44925216a6bc5f9c09cfae0">20cb17</a> AdvLoggerPkg: Add PanicLib instance (<a href="https://github.com/microsoft/mu_plus/pull/348">#348</a>)</li>
<li><a href="https://github.com/microsoft/mu_plus/commit/22011420bde4e643ece940aeede894df021144b0">220114</a> Add HID Keyboard support to UefiHidDxe (<a href="https://github.com/microsoft/mu_plus/pull/347">#347</a>)</li>
</ul>
</details>

Signed-off-by: Project Mu Bot <mubot@microsoft.com>

PR #767, opened 2023-11-09
Cargo.toml: Optimize dev profile binary size

makubacki:
## Description

By default, the dev profile is used. The default build settings for the dev profile are documented here:
https://doc.rust-lang.org/cargo/reference/profiles.html#dev

Unmodified dev profile settings result in extraordinarily large binaries relative to UEFI FW. This especially impacts `DEBUG` builds which already have less optimized C code resulting in overall greater space occupation.

Without a change, the binaries are simply too large and will continue to push the limits of firmware volumes (on a real system constrained by flash size) over time.

Therefore, the below setting enables optimization level `3` (all optimizations) that is used by the release profile by default. This greatly reduces the overall binary size. `[profile.dev.package."*"]` is specified to apply the opt-level for all dependencies (but not a workspace member). This emphasizes debuggability of workspace code but optimizes dependencies. An individual dependency can be overridden by specifying the named package instead of `"*"`. For example:

```ini
[profile.dev.package.foo]
opt-level = 0
```

That will likely allow the overall build to still fit in the FV, since other code is still optimized, but remove optimizations from an individual package that needs to be debugged.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified build size of several config combinations (including the chosen one).
  - ```
    [profile.dev]
    opt-level = 'z'
    ```
    - DXE FV: 11643648 (0xb1ab00) used (size opt of all packages)
  - ```
    [profile.dev.package."*"]
    opt-level = 'z'
    ```
    - DXE FV: 12365056 (0xbcad00) used (size opt of dependencies)
  - ```
    [profile.dev.package."*"]
    opt-level = 3
    ```
    - DXE FV: 12431104 (0xbdaf00) used (level 3 opt of dependencies)

## Integration Instructions

N/A - Affects packages built in this workspace

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,25 @@ memoffset = "0.9.0"
 r-efi = "4.3.0"
 rustversion = "1.0.14"
 spin = "0.5.2"
+
+# By default, the dev profile is used. The default build settings for the dev profile are documented here:
+# https://doc.rust-lang.org/cargo/reference/profiles.html#dev
+#
+# Unmodified dev profile settings result in extraordinarily large binaries relative to UEFI FW. This especially
+# impacts DEBUG builds which already have less optimized C code resulting in overall greater space occupation.
+# Without a change, the binaries are simply too large and will continue to push the limits of firmware volumes
+# (on a real system constrained by flash size) over time.
+#
+# Therefore, the below setting enables optimization level 3 (all optimizations) that is used by the release profile
+# by default. This greatly reduces the overall binary size. [profile.dev.package."*"] is specified to apply the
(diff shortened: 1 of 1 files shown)


makubacki:
PR #764 is blocked due to DXEFV size w/ GCC. Verified 151KB will still be available in DXEFV GCC build with that change after this PR.

Javagedes (approved):

TaylorBeebe (approved):

PR #851, opened 2024-02-07
Use qemu external dependency

Javagedes:
## Description

Use an external dependency for qemu rather than relying on it being installed by the user.

- [ ] Impacts functionality?
  - **Functionality** - Does the change ultimately impact how firmware functions?
  - Examples: Add a new library, publish a new PPI, update an algorithm, ...
- [ ] Impacts security?
  - **Security** - Does the change have a direct security impact on an application,
    flow, or firmware?
  - Examples: Crypto algorithm change, buffer overflow fix, parameter
    validation improvement, ...
- [ ] Breaking change?
  - **Breaking change** - Will anyone consuming this change experience a break
    in build or boot behavior?
  - Examples: Add a new library class, move a module to a different repo, call
    a function in a new library class in a pre-existing module, ...
- [ ] Includes tests?
  - **Tests** - Does the change include any explicit test code?
  - Examples: Unit tests, integration tests, robot tests, ...
- [ ] Includes documentation?
  - **Documentation** - Does the change contain explicit documentation additions
    outside direct code modifications (and comments)?
  - Examples: Update readme file, add feature readme file, link to documentation
    on an a separate Web page, ...

## How This Was Tested

Ensured Boot to shell and Boot to windows using the Windows and Linux qemu external dependencies. Verified TPM Replay functionality continued to work as expected.

## Integration Instructions

N/A

PR #852, opened 2024-02-08
Add qemu binary release workflow:

Javagedes:
## Description

Add a new workflow that will build and prepare qemu binaries for both Windows and Linux. This workflow has three triggers:

1. A pull request to the main branch, which will only trigger when either the workflow itself, or the qemu version file is updated. This workflow will upload the binaries as an artifact to the workflow.

2. A release, which will upload the binaries as an artifact to the associated release

3. A manual trigger, which allows for testing the workflow on a custom branch. This will upload the binaries as an artifact to the workflow.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

A release was performed using this github action on a personal branch ([v0.1.9](https://github.com/Javagedes/mu_tiano_platforms/releases/tag/v0.1.9)), and was used as an external dependency on a mu_tiano_platforms [Draft PR](https://github.com/microsoft/mu_tiano_platforms/pull/851), which is passing, and will be merged once this PR has been merged and a release has been executed.

## Integration Instructions

N/A


Javagedes:
@kuqin12 , @cfernald, Do you have some additional use cases that I / You can test? Probably related to loading an OS - but anything important?

makubacki:
> @kuqin12 , @cfernald, Do you have some additional use cases that I / You can test? Probably related to loading an OS - but anything important?

If you have time to test the TpmReplay feature, instructions are here - https://github.com/microsoft/mu_tiano_platforms/blob/main/Platforms/Docs/Q35/Features/feature_tpm_replay.md. This is a little unique because it uses a TPM emulator. The TPM emulator only works on Linux at the moment. If there is not time to test it, it is used rarely enough, I can take a note to check it later.

I recommend at least booting Windows if you haven't done that.

kuqin12:
> > @kuqin12 , @cfernald, Do you have some additional use cases that I / You can test? Probably related to loading an OS - but anything important?
> 
> If you have time to test the TpmReplay feature, instructions are here - https://github.com/microsoft/mu_tiano_platforms/blob/main/Platforms/Docs/Q35/Features/feature_tpm_replay.md. This is a little unique because it uses a TPM emulator. The TPM emulator only works on Linux at the moment. If there is not time to test it, it is used rarely enough, I can take a note to check it later.
> 
> I recommend at least booting Windows if you haven't done that.

Agreed. We normally would validate with a Windows boot with the QEMU UEFI.

Javagedes:
> > > @kuqin12 , @cfernald, Do you have some additional use cases that I / You can test? Probably related to loading an OS - but anything important?
> > 
> > 
> > If you have time to test the TpmReplay feature, instructions are here - https://github.com/microsoft/mu_tiano_platforms/blob/main/Platforms/Docs/Q35/Features/feature_tpm_replay.md. This is a little unique because it uses a TPM emulator. The TPM emulator only works on Linux at the moment. If there is not time to test it, it is used rarely enough, I can take a note to check it later.
> > I recommend at least booting Windows if you haven't done that.
> 
> Agreed. We normally would validate with a Windows boot with the QEMU UEFI.

Thanks! I will test both.

kuqin12 (approved):

.github/workflows/publish-qemu-bin.yml
```diff
@@ -0,0 +1,163 @@
...
+      run: |
+        $qemuVersion = (Get-Content -Path $env:QEMU_VERSION_FILE | ConvertFrom-Json).windows
+        "QEMU_VERSION=$qemuVersion" | Out-File -FilePath "$env:GITHUB_OUTPUT" -Append
+      shell: pwsh
+    - name: Download Qemu v${{ steps.qemu-version.outputs.QEMU_VERSION }}
+      env:
+        QEMU_VERSION: ${{ steps.qemu-version.outputs.QEMU_VERSION }}
+      run: |
+        choco install qemu --version=$env:QEMU_VERSION
+    - name: Stage Qemu Binaries
```
makubacki, line 36: Did you check if any license files need to be included for any of the binary components?

Kind of like the [license file](https://github.com/microsoft/mu_crypto_release/blob/release/202302/CryptoBinPkg/Driver/Packaging/License.txt) included in shared crypto releases.

makubacki (approved):

.github/workflows/publish-qemu-bin.yml
```diff
@@ -0,0 +1,163 @@
...
+          mv linux/ qemu/Linux/
+          ls qemu/Windows
+          ls qemu/Linux
+          chmod a+x qemu/Linux/*
+      - name: Package Ext-Dep
+        run: |
+          zip -r qemu-${{github.event.release.tag_name}}.zip qemu/*
+          tar -czf qemu-${{github.event.release.tag_name}}.tar.gz qemu/*
+
+      - name: Upload Ext-Dep (Artifact)
```
cfernald, line 149: If i understand this correctly, we will only be able to use release versions officially, right? That is there would be no way for a given release tag to use the qemu from its own release as an ext-dep right?

.github/workflows/publish-qemu-bin.yml
```diff
@@ -0,0 +1,163 @@
...
+          path: linux
+      - name: Stage Ext-Dep
+        run: |
+          mkdir -p qemu
+          mv windows/ qemu/Windows/
+          mv linux/ qemu/Linux/
+          ls qemu/Windows
+          ls qemu/Linux
+          chmod a+x qemu/Linux/*
+      - name: Package Ext-Dep
```
cfernald, line 144: I think it may make sense to just have the linux dir in .tar.gz and just the windows dir in the .zip. Not point downloading both in any given ext-dep, and it would cut the download in half.

.github/workflows/publish-qemu-bin.yml
```diff
@@ -0,0 +1,163 @@
...
+  
+  prepare-extdep:
+    name: Prepare and Upload Qemu External Dependency
+    needs: [qemu-windows, qemu-linux]
+    runs-on: Ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout
```
cfernald, line 121: Doesnt seem this checkout is actually used.

.github/workflows/publish-qemu-bin.yml
```diff
@@ -0,0 +1,163 @@
...
+        "QEMU_VERSION=$qemuVersion" | Out-File -FilePath "$env:GITHUB_OUTPUT" -Append
+      shell: pwsh
+    - name: Download Qemu v${{ steps.qemu-version.outputs.QEMU_VERSION }}
+      env:
+        QEMU_VERSION: ${{ steps.qemu-version.outputs.QEMU_VERSION }}
+      run: |
+        choco install qemu --version=$env:QEMU_VERSION
+    - name: Stage Qemu Binaries
+      run: |
+        mkdir temp
```
cfernald, line 38: Not for now, but we should eventually consider building this ourselves. It's a huge pain to build QEMU for windows, but might be useful as the existing one has some issues with things like WHPX. Not for this PR, but a note for the future.

.github/workflows/publish-qemu-bin.yml
Javagedes: Yes, that is correct - It is the same issue we face with BaseTools extdep in MU_BASECORE where only a commit *after* a release, could we start using the extdep from said release.

cfernald: Yeah, I recall that. I don't honest have another solution short of using nuget or something which isn't any better IMO.

Javagedes: One solution we could look at running the pipeline when each PR is committed to the default branch, and upload / replace the previous version in the release draft. Once the draft is published, it is locked in.

makubacki: I think the fundamental issue is that a version is being applied to two independent things - (1) firmware code for QEMU (2) QEMU itself.

Ideally, qemu should be released independently and only the version being consumed tracked in this repo.

makubacki: My understanding is we need a better replacement for NuGet feeds.

If we go with the GitHub release attachments the GitHub repo really needs to more closely represent the release (e.g. this repo should just be releasing firmware with the QEMU version tracked).

The workflow could be moved to Mu DevOps as a general catch all for flows like this but then using GitHub releases there would couple it with the Mu DevOps version which is also unrelated to QEMU. The Mu DevOps GitHub NuGet feed is available so the package could be independently versioned but we don't want to use NuGet.

In lieu of that, I was fine with making do here.

.github/workflows/publish-qemu-bin.yml
Javagedes: I will mark this down as a TODO in a future PR :)

.github/workflows/publish-qemu-bin.yml
Javagedes: Removed, thanks!

.github/workflows/publish-qemu-bin.yml
Javagedes: I've split them into separate external dependencies:

`qemu-windows-$(release-version).zip`
`qemu-linux-$(release-version).tar.gz`

.github/workflows/publish-qemu-bin.yml
Javagedes: Just kidding. Had to add it back in - but now it's being used.

.github/workflows/publish-qemu-bin.yml
Javagedes: We now upload

- https://github.com/qemu/qemu/blob/master/COPYING
- https://github.com/qemu/qemu/blob/master/LICENSE

Using the specific tag we are downloading - not master.

Javagedes:
Verified Tpm Replay works.

PR #867, opened 2024-02-28
QemuQ35Pkg: Update PlatformSecureLib to Qemu PlatformSecureLib

Flickdm:
## Description

Functionally, this just changes PlatformSecureLib to always return TRUE
So that UserPhysicalPresent(..) returns True

- [X] Impacts functionality?
  - **Functionality** - UserPhysicalPresent(..) always returns True
- [ ] Impacts security?
  - **Security** - No
- [ ] Breaking change?
  - **Breaking change** - No
- [ ] Includes tests?
  - **Tests** - No
- [ ] Includes documentation?
  - **Documentation** - No

## How This Was Tested

Tested locally to confirm UserPhysicalPresent(..) returns true when attempting to delete a secure boot authenticated variable.

## Integration Instructions

N/A


Flickdm:
After talking to Sean, UserPhysicalPresent(..) should never return true in Project MU

PR #891, opened 2024-03-27
[REBASE & FF] Wait for Serial Connection When SERIAL_PORT is Set in Q35 and SBSA

os-d:
## Description

Removing 'nowait' from the QEMU command line will prompt QEMU to wait to execute the FW until a serial connection is made. This is valuable for debugging FW and avoids a developer needing to quickly enable a debugger.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted with GDB_SERVER and saw it wait.

## Integration Instructions

If GDB_SERVER is specified, connect the GDB server before QEMU execution continues.

### DIFF
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -218,9 +218,10 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
             args += " -gdb tcp::" + gdb_port
 
         # write ConOut messages to telnet localhost port
+        # Force QEMU to wait for serial connection before booting
         serial_port = env.GetValue("SERIAL_PORT")
         if serial_port != None:
-            args += " -serial tcp:127.0.0.1:" + serial_port + ",server,nowait"
+            args += " -serial tcp:127.0.0.1:" + serial_port + ",server"
 
         # Connect the debug monitor to a telnet localhost port
         monitor_port = env.GetValue("MONITOR_PORT")
--- a/Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
@@ -131,9 +131,10 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
             args += " -gdb tcp::" + gdb_port
 
         # write ConOut messages to telnet localhost port
+        # Force QEMU to wait for serial connection before booting
         serial_port = env.GetValue("SERIAL_PORT")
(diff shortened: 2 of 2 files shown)


Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -211,11 +211,11 @@ def Runner(env):
         else:
             args += " -vga cirrus" #std is what the default is
 
-        # Check for gdb server setting
+        # Check for gdb server setting and wait for a connection
         gdb_port = env.GetValue("GDB_SERVER")
         if (gdb_port != None):
             logging.log(logging.INFO, "Enabling GDB server at port tcp::" + gdb_port + ".")
-            args += " -gdb tcp::" + gdb_port
+            args += " -S -gdb tcp::" + gdb_port
```
cfernald, line 218: I actually like that it doesnt wait, most of the time. I'm curious, how to you break iin on anything meaningful just by catching it earlier?

os-d: I was just having this chat with Kun. I guess in this case waiting for the serial port to be connected is more important than the GDB server. That's where you can miss an OS debugger if you didn't get the serial port connected in time. Yes you typically have awhile but also whenever I debug I'm multitasking and I miss it. And since the serial logs dumps to the terminal and the file, I don't have any use case for adding a serial connection otherwise. So I can change this PR to do that, if people are interested in it. I've been running intermixed with both of these debugging options and I agree that for the EXDI debugging you have to put a deadloop anyway, so forcing it to wait probably doesn't matter and my real case is the serial debugging.

cfernald: Do you not just start qemu and windbg at the same time? Sorry, I'm not attacking your scenario, mostly just want to understand if there is opportunity for improvement as we move to a more debuggable QEMU and UEFI.

os-d: I do, but when I'm multitasking sometimes I forget :). There is also the kgdb case for debugging Linux and I don't recall if gdb waits for the connection.

TBH, I put up this PR because I said, I keep adding these QemuRunner flags, let's just put them upstream. But honestly, if there are better flows I'm fine to drop these.

os-d: @cfernald, for serial debugging over TCP you can't have WinDbg open until QEMU starts and creates the tcp connection, though, right? So this does help for that case.

PR #900, opened 2024-04-08
Fix console after returning from QEMU

kuqin12:
# Preface

Please ensure you have read the [contribution docs](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md) prior
to submitting the pull request. In particular,
[pull request guidelines](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md#pull-request-best-practices).

## Description

The fix is enlightened by the original QEMU thread here: https://gitlab.com/qemu-project/qemu/-/issues/1674.

The change fixed the garbled command prompt after returning from QEMU v8+ on Windows platforms.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU v9.0.0-rc1 on Windows and confirmed it fixes the command prompt after returning from QEMU runner.

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -227,11 +227,14 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
         if monitor_port is not None:
             args += " -monitor tcp:127.0.0.1:" + monitor_port + ",server,nowait"
 
+        ## TODO: Save the console mode. The original issue comes from: https://gitlab.com/qemu-project/qemu/-/issues/1674
+        if os.name == 'nt' and qemu_version[0] >= '8':
+            import win32console
+            std_handle = win32console.GetStdHandle(win32console.STD_INPUT_HANDLE)
+            console_mode = std_handle.GetConsoleMode()
+
         # Run QEMU
         ret = utility_functions.RunCmd(executable, args)
-        if ret != 0 and os.name != 'nt':
-            # Linux version of QEMU will mess with the print if its run failed, this is to restore it
-            utility_functions.RunCmd ('stty', 'echo')
 
         ## TODO: restore the customized RunCmd once unit tests with asserts are figured out
         if ret == 0xc0000005:
@@ -243,12 +246,11 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
-        ## TODO: Recover the console. Remove this once QEMU fixes the issue: https://gitlab.com/qemu-project/qemu/-/issues/1674
+        ## TODO: Save the console mode. The original issue comes from: https://gitlab.com/qemu-project/qemu/-/issues/1674
         if os.name == 'nt' and qemu_version[0] >= '8':
-            import win32console
-            h = win32console.GetStdHandle(win32console.STD_INPUT_HANDLE)
-            mode = h.GetConsoleMode()
-            # pywin32 does not have the constants for ENABLE_VIRTUAL_TERMINAL_INPUT
-            h.SetConsoleMode(mode & ~0x0200)
+            std_handle.SetConsoleMode(console_mode)
+        elif os.name != 'nt':
+            # Linux version of QEMU will mess with the print if its run failed, let's just restore it anyway
+            utility_functions.RunCmd ('stty', 'sane')
 
         return ret
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -243,4 +243,12 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
+        ## TODO: Recover the console. Remove this once QEMU fixes the issue: https://gitlab.com/qemu-project/qemu/-/issues/1674
+        if os.name == 'nt' and qemu_version[0] >= '8':
+            import win32console
+            h = win32console.GetStdHandle(win32console.STD_INPUT_HANDLE)
+            mode = h.GetConsoleMode()
+            # pywin32 does not have the constants for ENABLE_VIRTUAL_TERMINAL_INPUT
+            h.SetConsoleMode(mode & ~0x0200)
+
         return ret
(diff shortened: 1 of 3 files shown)


Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -243,4 +243,12 @@ def Runner(env):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
+        ## TODO: Recover the console. Remove this once QEMU fixes the issue: https://gitlab.com/qemu-project/qemu/-/issues/1674
+        if os.name == 'nt' and qemu_version[0] >= '8':
```
os-d, line 247: Is it worthwhile to do the equivalent on Linux, just in case? `stty sane` will  reset most things.

os-d (approved):

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
kuqin12: I did have some handling for that around line 234, but that was only for when Linux instance returned failure. Do we really want to apply stty change to all returns?

os-d: I think I've only seen the Linux terminal get messed up on qemu failure, I was more just thinking of doing a cleanup post qemu, since it does mess with terminal stuff. But I don't care too much one way or another

kuqin12: Okay, modified the flow so that it will restore to the original console mode for Windows and `stty sane` for Linux.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -227,11 +227,14 @@ def Runner(env):
         if monitor_port is not None:
             args += " -monitor tcp:127.0.0.1:" + monitor_port + ",server,nowait"
 
+        ## TODO: Save the console mode. The original issue comes from: https://gitlab.com/qemu-project/qemu/-/issues/1674
+        if os.name == 'nt' and qemu_version[0] >= '8':
+            import win32console
```
Javagedes, line 232: win32console is a part of pywin32, so it needs to be added to the pip requirements:

Example: <https://github.com/microsoft/mu_feature_config/blob/main/pip-requirements.txt?

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -243,4 +246,11 @@ def Runner(env):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
+        if os.name == 'nt' and qemu_version[0] >= '8':
+            # Restore the console mode for Windows on QEMU v8+.
+            std_handle.SetConsoleMode(console_mode)
+        elif os.name != 'nt':
```
Javagedes, line 252: Should this be elif os.name != 'nt' and qemu_verion[0] >= '8'?

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -243,4 +246,11 @@ def Runner(env):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
+        if os.name == 'nt' and qemu_version[0] >= '8':
```
Javagedes, line 249: Does qemu 9.0.0 fix this? if so is it more appropriate to do == '8'?

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -227,11 +227,14 @@ def Runner(env):
         if monitor_port is not None:
```
Javagedes, line 227: Do these changes need to be mirrored to QemuSbsa's QemuRunner? At a minimum it seems like the `stty sane` would need to be done there too.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -243,4 +246,11 @@ def Runner(env):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
+        if os.name == 'nt' and qemu_version[0] >= '8':
+            # Restore the console mode for Windows on QEMU v8+.
+            std_handle.SetConsoleMode(console_mode)
+        elif os.name != 'nt':
+            # Linux version of QEMU will mess with the print if its run failed, let's just restore it anyway
+            utility_functions.RunCmd ('stty', 'sane')
```
Javagedes, line 254: Please either set `logging_level = logging.DEBUG` or `capture=false` to stop this from spitting out junk to the terminal post build.

Javagedes (changes requested):

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
kuqin12: Will do.

kuqin12: Updated.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
kuqin12: Updated with `capture=false`

Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -159,4 +162,11 @@ def Runner(env):
             # Tested same FDs on QEMU 6 and 7, not observing the same.
             ret = 0
 
+        if os.name == 'nt' and qemu_version[0] >= '8':
+            # Restore the console mode for Windows on QEMU v8+.
+            std_handle.SetConsoleMode(console_mode)
+        elif os.name != 'nt':
```
Javagedes, line 168: Does this need the same >= '8' condition?

Javagedes (approved):

Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
kuqin12: Not really, the Linux versions have been giving trouble to the runner console for more than a year (if not 2 years). We definitely want to restore the `stty` state when the return code is incorrect, but I agree with Oliver that a clean up seems to be needed regardless as QEMU kept polluting the returned console here and there. Let me know if you have any concerns.

PR #1036, opened 2024-09-13
Enable Debugger By Default in Q35/SBSA

os-d:
## Description

This PR turns mu_feature_debugger on by default in Q35/SBSA. It sets no initial breakpoint, but will break in on an exception. As part of this, QemuRunner defaults to having `SERIAL_PORT=50001` if this is not overridden, so that on a regular boot of Q35 the debugger can easily be used. On SBSA, the serial port is not added by default, as it will prevent writing to stdout, making it hard to tell progress.

An additional command line flag is added `BLD_*_DXE_DBG_BRK` to control whether the initial breakpoint is set. Passing `BLD_*_DXE_DBG_BRK=TRUE` on the commandline will set an unlimited initial breakpoint on DxeCore entrance.

The other caveat in SBSA is that with a single serial port, debugger polling is disabled, otherwise the debugger can eat user keystrokes at the shell.

Once supervised StMM support is added to mu_feature_debugger (https://github.com/microsoft/mu_feature_debugger/issues/59), it will be added here with a corresponding flag to force breaking in to the MM debugger.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting Q35 with no exceptions, with an exception and breaking in, and with the force flag and breaking in on initial breakpoint.

## Integration Instructions

See above.

### DIFF
--- a/Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
+++ b/Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
@@ -737,7 +737,13 @@
   # Bit 1 - Controls whether the DXE debugger is enabled.
   # Bit 2 - Controls whether the MM debugger is enabled.
   # Bit 3 - Disables the debuggers periodic polling for a requested break-in.
+  # if BLD_*_DXE_DBG_BRK=TRUE, the debugger will have an initial break in, but will break in, otherwise it only will
+  # on exceptions
+!if $(DXE_DBG_BRK) == TRUE
+  DebuggerFeaturePkgTokenSpaceGuid.PcdDebugConfigFlags|0x3
+!else
   DebuggerFeaturePkgTokenSpaceGuid.PcdDebugConfigFlags|0x2
+!endif
 
   # Set the debugger timeout to wait forever. This only takes effect if Bit 0 of PcdDebugConfigFlags is set
   # to 1, which by default it is not. Using BLD_*_DXE_DBG_BRK=TRUE will set this to 1.
@@ -1032,13 +1038,6 @@ QemuQ35Pkg/Library/ResetSystemLib/StandaloneMmResetSystemLib.inf
     <LibraryClasses>
       NULL|MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf
       DevicePathLib|MdePkg/Library/UefiDevicePathLib/UefiDevicePathLib.inf
-    <PcdsFixedAtBuild>
-      ## Forcibly enables the debugger with a 30 second initial breakpoint. This can be set from the cmdline by passing
-      # BLD_*_DXE_DBG_BRK=TRUE. If this is false, the debugger will not have an initial break in, but will break in
-      # on exceptions
-      !if $(DXE_DBG_BRK) == TRUE
-        DebuggerFeaturePkgTokenSpaceGuid.PcdDebugConfigFlags|0x3
-      !endif
   }
 
   MdeModulePkg/Universal/ReportStatusCodeRouter/RuntimeDxe/ReportStatusCodeRouterRuntimeDxe.inf
--- a/Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
+++ b/Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
@@ -848,12 +848,13 @@
   gEfiMdePkgTokenSpaceGuid.PcdConfidentialComputingGuestAttr|0
 
   # Add DEVICE_STATE_UNIT_TEST_MODE to the device state bitmask if BUILD_UNIT_TESTS=TRUE (default)
+  # in addition to debugger enabled
   !if $(BUILD_UNIT_TESTS) == TRUE
-    gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x20
-  !endif
-
+    gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x28
+  !else
   # Set to debug as debugger is enabled.
-  gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x08
+    gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x08
+  !endif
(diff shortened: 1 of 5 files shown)


kuqin12 (approved):

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
```diff
@@ -1035,6 +1037,11 @@ QemuQ35Pkg/Library/ResetSystemLib/StandaloneMmResetSystemLib.inf
     <LibraryClasses>
       NULL|MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf
       DevicePathLib|MdePkg/Library/UefiDevicePathLib/UefiDevicePathLib.inf
+    <PcdsFixedAtBuild>
+      ## Forcibly enables the debugger with a 30 second initial breakpoint. This can be set from the cmdline by passing
```
cfernald, line 1041: Force enabled will enabled a inifinite timeout.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
```diff
@@ -29,8 +29,8 @@
   # Defines for default states.  These can be changed on the command line.
   # -D FLAG=VALUE
   #
-!ifndef DEBUGGER_ENABLED
-  DEFINE DEBUGGER_ENABLED               = FALSE
+!ifndef FORCE_DXE_DEBUGGER
```
cfernald, line 32: What is the benefit of this? It seems if we have the configuration pei module there is no need to force enable.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
```diff
@@ -840,10 +844,8 @@
     gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x20
   !endif
 
-  # Set to debug if debugger is enabled.
-  !if $(DEBUGGER_ENABLED) == TRUE
-    gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x08
-  !endif
+  # Set to debug as debugger is enabled.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdDeviceStateBitmask|0x08
```
cfernald, line 848: This will cause the default 30 second timeout. Do we want to drop that down to like 5 seconds or something to not slow down pipelines, test, etc?

makubacki: Does it make sense to add a build flag to not set this that can be used in cases like pipelines?

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
os-d: Yep, I see that now, misread the hob def.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
os-d: Yeah, agreed. I will set this to control the initial breakpoint timeout only.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
os-d: @cfernald how will this cause the 30 second timeout? I thought this just controls whether the debugger can be enabled or not: https://github.com/microsoft/mu_feature_debugger/blob/24f5ebe0782b35e13e1529c405f94f9d776feae9/DebuggerFeaturePkg/DebugConfigPei/DebugConfigPei.c#L42.

cfernald: Yes, but sense you do not specify the PCD's you will use the defaults: https://github.com/cfernald/mu_feature_debugger/blob/62f972757843a1c9b583efd681ddac6f65ac11bf/DebuggerFeaturePkg/DebuggerFeaturePkg.dec#L52

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
cfernald: I think what you want is to leave the timeout alone and have a flag to control whether you get the initial breakpoint, by seeing bit 0 of the flags control:  https://github.com/cfernald/mu_feature_debugger/blob/62f972757843a1c9b583efd681ddac6f65ac11bf/DebuggerFeaturePkg/DebuggerFeaturePkg.dec#L43

os-d: Yep agreed, that's what I meant, but typed it out wrong :)

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
os-d: @cfernald , but I am specifying the `PcdDebugConfigFlags` saying no initial breakpoint, so it doesn't matter what the default is. So with the other comment, I will change this to specifying timeout = 0 always, but the initial breakpoint is off, unless you set the force flag. Then CI is not affected as by default no initial breakpoint, if you are overriding from the commandline, that is you saying yes, I want to break in, so wait for me forever. If the cli override is not set, you will not have an initial break in and will only break in if an exception occurs.

cfernald: Ah, sorry i missed that line in my original pass. you are right.

Platforms/QemuQ35Pkg/QemuQ35Pkg.dsc
cfernald: ah, gotcha. So we wouldn't be using the force enable then, right?

os-d: Yeah, switched off of force enable.

cfernald (approved):

apop5 (approved):

makubacki (approved):

PR #1051, opened 2024-10-03
QemuSbsaPkg: QemuRunner: Don't Expose Default Serial Port

os-d:
## Description

This was PR feedback that was missed. SBSA only has a single serial port, so if we put out one by default, it will not print to stdio and we will miss the logs.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.


apop5 (approved):

cfernald (approved):

apop5:
duplicate of #1057, which has merged.

PR #1068, opened 2024-11-15
FF-A SP and Service Changes (Notification, TPM, Test):

Raymond-MS:
## Description

Updated the mu_tiano_platforms build (ffa_enablement) to work with the latest ms-services SP including: Notification, Test, and TPM. PlatformBuild.py currently points to a hard coded SP binary but will be updated when we port the SP and services to UEFI style. Updated the TestApp to test all services.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Functionality tested with FfaPartitionTestApp and running with TPM enabled

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuSbsaPkg/FfaPartitionTest/FfaPartitionTestApp.c
+++ b/Platforms/QemuSbsaPkg/FfaPartitionTest/FfaPartitionTestApp.c
@@ -29,10 +29,21 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/UefiLib.h>
 #include <Library/UefiBootServicesTableLib.h>
 
+#define FFA_NOTIFICATION_SERVICE_GUID \
+  { \
+    0xb510b3a3, 0x59f6, 0x4054, { 0xba, 0x7a, 0xff, 0x2e, 0xb1, 0xea, 0xc7, 0x65 } \
+  }
+
+#define FFA_TPM_SERVICE_GUID \
+  { \
+    0x17b862a4, 0x1806, 0x4faf, { 0x86, 0xb3, 0x08, 0x9a, 0x58, 0x35, 0x38, 0x61 } \
+  }
+
 #define FFA_TEST_SERVICE_GUID \
   { \
-    0x7036A3DB, 0xBA3B, 0x44E5, { 0xB3, 0x59, 0xE2, 0x5E, 0x5D, 0x90, 0x43, 0x87 } \
+    0xe0fad9b3, 0x7f5c, 0x42c5, { 0xb2, 0xee, 0xb7, 0xa8, 0x23, 0x13, 0xcd, 0xb2 } \
   }
+
 
 UINT16  FfaPartId;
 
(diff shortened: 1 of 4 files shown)


kuqin12 (approved):

Platforms/QemuSbsaPkg/PlatformBuild.py
```diff
@@ -391,14 +391,14 @@ def PlatformPostBuild(self):
                 },
                 "example": {
                     "image": {
-                        "file": os.path.join(self.env.GetValue('BUILD_OUTPUT_BASE'), 'FV', 'BL32_AP_MM_SP1.fd'),
+                        "file": "/home/rdiaz/Projects/trusted-services/deployments/ms-services/config/default-sp/build/b8bcbd0c-8e8f-4ebe-99eb-3cbbdd0cd412.bin",
```
makubacki, line 394: You'll need to get rid of the user-local path before merging.

Platforms/QemuSbsaPkg/FfaPartitionTest/FfaPartitionTestApp.c
```diff
@@ -209,42 +221,112 @@ FfaPartitionTestAppEntry (
...
+    DEBUG ((DEBUG_ERROR, "Unable to communicate direct req 2 with FF-A Ffa test SP (%r).\n", Status));
     goto Done;
   }
 
-  DUMP_HEX (DEBUG_INFO, 0, &DirectMsgArgs, sizeof (DirectMsgArgs), "    ");
+  if (DirectMsgArgsEx.Arg0 != 0) {
+    DEBUG ((DEBUG_ERROR, "Command Failed: %x\n", DirectMsgArgsEx.Arg0));
+    goto Done;
+  } else {
+    DEBUG ((DEBUG_ERROR, "Thermal Service Setup Success\n"));
```
makubacki, line 249: Was there a particular reason to print these success messages at error level?

```suggestion
    DEBUG ((DEBUG_INFO, "Thermal Service Setup Success\n"));
```

makubacki (approved):

Platforms/QemuSbsaPkg/PlatformBuild.py
Raymond-MS: No worries, I am aware that this needs to change. There are plans to generate the SP using UEFI instead of Trusted Services which is where this binary comes from.

Platforms/QemuSbsaPkg/FfaPartitionTest/FfaPartitionTestApp.c
Raymond-MS: No, there was no reason, it's mainly just a copy-paste thing. The entire TestApp is just for verifying the SP functionality and won't ever be used outside of that, as far as I understand. I just wanted print statements to verify things were working as intended. :)

makubacki: That's fine. It is still nice to stick to the expected levels so it matches people's expectations and you'd be surprised where this could end up getting copied/pasted.

Platforms/QemuSbsaPkg/PlatformBuild.py
makubacki: Sounds good. It is okay to have a temporary path, but it should be relative to the workspace or if it is truly an example where the explicitness of an absolute path helps, anonymize it.

Platforms/QemuSbsaPkg/FfaPartitionTest/FfaPartitionTestApp.c
Raymond-MS: Agreed. I will update. Looks like I had INFO in some places but not all. That is my fault.

PR #1101, opened 2025-01-15
[FFA][TPM] Fix TPM2_ENABLE flag

kuqin12:
## Description

This change updates the TPM flag to guard the firmware from communicating to TPM devices.

This is needed so that it does not require all QEMU foundation to be updated.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on QEMU SBSA platform.

## Integration Instructions

N/A


Raymond-MS (approved):
Looks good to me, however, it looks like you need to run Uncrustify. The CI isn't passing.

PR #1102, opened 2025-01-21
[FFA] Fixed an issue of symbol unresolved when enabling TPM

kuqin12:
## Description

The `TpmServiceLib` is not linkable when there is library class specified because its inf does not pertain the same logic of linking `TpmServiceLib`.

This change moved it to NULL library class so that it will be linked according to the build flag.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested buildable for QEMU SBSA.

## Integration Instructions

N/A


Raymond-MS (approved):
Looks good. Thanks for fixing!

apop5 (approved):

kuqin12:
> Looks good. Thanks for fixing!

thanks for the review. sorry for the hassle :)

PR #1105, opened 2025-01-30
use external dependency for cargo tools on windows

Javagedes:
## Description

Cargo tools (tarpaulin / make) are provided for free when using the provided devcontainer. However, on Windows, these tools must be installed or downloaded manually at a specific version. This has caused frequent issues for developers as local installation via compilation tend to fail as this project uses an outdated version of rust.

This change adds these tools as external dependencies, so that developers who only care about compiling the project can do so without having to worry about installing these additional tools.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Removed my installation of cargo-make and cargo-tarpaulin and was still able to run stuart_build and stuart_ci_build

## Integration Instructions

N/A

### DIFF
--- a/QemuPkg/Binaries/cargo-make_ext_dep.yaml
+++ b/QemuPkg/Binaries/cargo-make_ext_dep.yaml
@@ -1,0 +1,18 @@
+##
+# Downloads the exact version of cargo-make. This version should be the same version as that found in the
+# rust-toolchain.toml file.
+#
+# Copyright (c) 2019, Microsoft Corporation
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+{
+  "scope": "global-win",
+  "type": "web",
+  "name": "cargo-make",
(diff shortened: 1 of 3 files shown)


apop5:
I like this.

I am worried about another subhook situation, where the ext_dep source disappears.

Javagedes:
> I like this.
> 
> I am worried about another subhook situation, where the ext_dep source disappears.

This is definitely a fair point. we could publish them somewhere ourselves, Maybe as an ado artifact for project mu:

https://dev.azure.com/projectmu/mu/_artifacts/feed/Mu-Public

kuqin12 (approved):

makubacki:
> > I like this.
> > I am worried about another subhook situation, where the ext_dep source disappears.
> 
> This is definitely a fair point. we could publish them somewhere ourselves, Maybe as an ado artifact for project mu:
> 
> https://dev.azure.com/projectmu/mu/_artifacts/feed/Mu-Public

That sounds fine.

PR #1132, opened 2025-03-25
Resolve poetry failure on a secondary repo

kuqin12:
## Description

This change intends to setup poetry environments for the TFA build process.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change is tested locally.

## Integration Instructions

N/A


Platforms/QemuSbsaPkg/PlatformBuild.py
```diff
@@ -321,6 +321,27 @@ def SetPlatformEnvAfterTarget(self):
...
+        # If we are not in a virtual environment, we can build the firmware directly.
+        ret = RunCmd("poetry", "env list --full-path", workingdir=self.env.GetValue("ARM_TFA_PATH"), outstream=outstream, environ=cached_enivron)
+        if ret != 0:
+            # it is not setup
+            need_setup = True
+        elif outstream.getvalue() == "":
+            # it is not setup
+            need_setup = True
+
+        need_setup = True
```
Raymond-MS, line 337: Is there a reason to set this to True here always instead of letting it be set by the if/elif above? One of these should be removed, right?

kuqin12: Yeah.. Sorry, debug residue

Raymond-MS:
I don't feel super comfortable signing off on these changes as I don't understand them well enough. Would it be okay if you pulled in another reviewer to take a look as well? As far as I can tell the changes look okay but I don't have the experience in this to confirm there are no issues.

Raymond-MS (approved):
LGTM, but I would like another pair of eyes if possible. :)

apop5 (approved):

PR #1143, opened 2025-05-13
Fix calls to retrieve Cpu Count and Mpidr to use ArmMonitorCalls instead

eeshanl:
## Description

Fix calls to retrieve Cpu Count and Mpidr to use ArmMonitorCalls instead
Bump TFA to 2.13-rc

This fix is needed for newer qemu versions > 9

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU SBSA mu_tiano_platforms

## Integration Instructions

Requires TFA commit: https://github.com/ARM-software/arm-trusted-firmware/commit/42925c15bee09162c6dfc8c2204843ffac6201c1#diff-efe8a973d827b75aa34a8b6fd065bc9a7ffd33290d4a73697797e24e56460ae2R76)

### DIFF
--- a/Platforms/QemuSbsaPkg/SbsaQemuAcpiDxe/SbsaQemuAcpiDxe.c
+++ b/Platforms/QemuSbsaPkg/SbsaQemuAcpiDxe/SbsaQemuAcpiDxe.c
@@ -9,6 +9,8 @@
 #include <IndustryStandard/Acpi.h>
 #include <IndustryStandard/AcpiAml.h>
 #include <IndustryStandard/SbsaQemuAcpi.h>
+#include <IndustryStandard/ArmStdSmc.h>
+#include <Library/ArmMonitorLib.h>
 #include <Library/AcpiLib.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/DebugLib.h>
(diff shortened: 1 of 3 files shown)


kuqin12:
As you mentioned in the description, can we also update TFA to a release tag? 2.13-rc is just released, probably can just take that.

makubacki:
My understanding is that this should be marked "functional" and "type:bug". Is that correct? If so, can you please update the labels?

kuqin12 (approved):

Platforms/QemuSbsaPkg/SbsaQemuAcpiDxe/SbsaQemuAcpiDxe.c
```diff
@@ -9,18 +9,80 @@
...
+  )
+{
+  ARM_MONITOR_ARGS  SmcArgs;
+
+  SmcArgs.Arg0 = SIP_SVC_GET_CPU_COUNT;
+  ArmMonitorCall (&SmcArgs);
+
+  if (SmcArgs.Arg0 != SMC_SIP_CALL_SUCCESS) {
+    DEBUG ((DEBUG_ERROR, "%a: SIP_SVC_GET_CPU_COUNT call failed. We have no cpu information.\n", __func__));
+    ResetShutdown ();
```
Raymond-MS, line 49: From my fairly quick check into this function it looks like there is a possibility that this returns. In that case, we should return immediately afterward instead of printing the message and returning ARG1, right? Same goes for the MPIDR call. I know in the normal case this should reset but, to be safe, we should handle the case where it is unsupported.

I would suggest putting a return 0 or, probably better, some kind of ASSERT afterwards. :)

Raymond-MS (approved):

kuqin12 (approved):

PR #1145, opened 2025-05-14
Track TFA patches and automatically apply during build

cfernald:
## Description

Creates a feature to track TFA patches locally while still using the official TFA releases. This is for **temporarily** carrying TF-A changes while we attempt to take them upstream.

For TF-A development, `PATCH_TFA=FALSE` should be used to avoid the changes being injected unexpectedly.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Local build and boot to shell

## Integration Instructions

N/A


.gitmodules
```diff
@@ -38,6 +38,6 @@
 [submodule "Silicon/Arm/HAF"]
 	path = Silicon/Arm/HAF
 	url = https://github.com/kuqin12/hafnium.git
-[submodule "Features/FFA"]
-	path = Features/FFA
-	url = https://github.com/microsoft/mu_feature_ffa.git
```
kuqin12, line 43: nit: what is changed here?

kuqin12 (approved):

os-d (commented):
I mean...this is pretty ugly :). But I understand the motivation. This is going to break clangpdb to move forward, right? So would rely on the fix to the TF-A build process?

.gitmodules
Raymond-MS: I was thinking the same thing! Haha.

Raymond-MS:
Is there any way we can mark this commit such that we know to come back and revert it once the changes get upstreamed? That is my only real comment. :) Other than that, it's good we keep up to date with the latest drops from TF-A. Hopefully we can get things upstreamed before this change starts having conflicts or issues with applying the patches.

Raymond-MS (approved):

.gitmodules
cfernald: Just line ending changes my compiler decided to make :)

PR #1218, opened 2025-09-11
Adding MP management test application for SBSA

kuqin12:
## Description

The existing QEMU SBSA has certain issues in the TF-A side that would prevent the system to launch multiple cores with proper MP support. Specifically due to the following flow:

- The setup places BL1, BL2, BL31, STMM code, and UEFI variables in secure flash to emulate physical platform behavior.
- During POR, TFA BL1 powers on all cores but boots only the primary core. The secondary APs enter an idle loop, polling the mailbox command.
 Note: The secondary APs execute in XIP mode from flash where BL1 resides.
- Once the primary core transitions to STMM, the UEFI flash driver attempts to write to the flash device, using this logic
- This write operation switches the flash to program mode, which inadvertently wakes the secondary cores from WFE.
- Since the flash is no longer in read mode, the secondary cores encounter data abort faults when trying to fetch instructions.

This change in TFA patch will move the AP cores to be held in BL31 context. The wfi routine inside is created to prevent the BSP to be waken up too easily.

The rest of the change is to integrate the MP management driver and test to be run on CI pipeline.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU SBSA and passed the test application.

## Integration Instructions

N/A


Raymond-MS (approved):
LGTM. However, I am not experienced enough to feel confident in my review. If possible, another dev should take a deeper look at the functionality to confirm it is doing what is expected.

os-d (approved):

PR #1242, opened 2025-10-15
Add key interrupt so that it will reset the console mode

kuqin12:
## Description

If QEMU is interrupted with `Ctrl + C`, the QEMU will leave the odd console mode enabled.

There used to be some logic handling the console mode restoration, but that did not work with key interrupt case.

This change is updated to fix that case.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified that QEMU interrupted by keyboard no longer has odd console mode after the change.

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -244,7 +244,7 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
         try:
             ret = utility_functions.RunCmd(executable, args)
         except KeyboardInterrupt:
-            logging.critical("QEMU run interrupted.")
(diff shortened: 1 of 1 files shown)


Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -241,7 +241,11 @@ def Runner(env):
                 std_handle = None
 
         # Run QEMU
-        ret = utility_functions.RunCmd(executable, args)
+        try:
+            ret = utility_functions.RunCmd(executable, args)
+        except KeyboardInterrupt:
+            logging.critical("QEMU run interrupted.")
```
Javagedes: nit:
```suggestion
            logging.critical("QEMU run interrupted by user (ctrl+c).")
```

makubacki (approved):

PR #1268, opened 2025-11-10
Fixing memory usage of QEMU SBSA

kuqin12:
## Description

This change fixed an issue where the memory available specified in QEMU Runner has to be fixed to 2GB.

With the integration of FF-A, the MM communication buffer exposed to STMM secure partition has to match what was subscribed in the partition manifest, in this case the PCD is set to match the value in the DTS.

In addition, the fix includes an update in mu_silicon_arm where the memory for MM communicate buffer is carved out from the system memory. Accordingly, the platform does not need to manipulate the system memory PCDs.

Lastly, the TF-A manifest needs to be updated to over-subscribe the memory available to the system.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested locally and booted to both UEFI shell with 2GB memory and Windows Desktop with 8GB.

## Integration Instructions

Update to latest mu_silicon_arm_tiano.


Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -104,16 +106,14 @@ def Runner(env):
...
-            args += f" -drive file={VirtualDrive},if=virtio"
-        elif os.path.isdir(VirtualDrive):
-            args += f" -drive file=fat:rw:{VirtualDrive},format=raw,media=disk"
         else:
-            logging.critical("Virtual Drive Path Invalid")
-
-        # TODO: Set the memory size to be 2GB regardless. Not sure why 8GB does
-        # not work.
-        args += " -m 2048"
+            args += " -m 2048"
```
os-d: Isn't this still setting 2048 always? it would just set 8192 first and then 2048 after?

os-d (approved):

PR #1270, opened 2025-11-11
Pick up new binaries of QEMU v10

kuqin12:
## Description

This change picks up the new binary releases of QEMU v10.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built QEMU Q35 firmware and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuSbsaPkg/Library/PlatformPeiLib/PlatformPeiLib.c
+++ b/Platforms/QemuSbsaPkg/Library/PlatformPeiLib/PlatformPeiLib.c
@@ -160,14 +160,9 @@ InitializeMemoryConfiguration (
     PANIC ("System Memory Base Mismatch.\n");
   }
 
-  PcdStatus = PcdSet64S (PcdSystemMemorySize, NewSize - PcdGet64 (PcdMmBufferSize));
+  PcdStatus = PcdSet64S (PcdSystemMemorySize, NewSize);
   if (EFI_ERROR (PcdStatus)) {
     PANIC ("Failed to set PcdSystemMemorySize.\n");
-  }
-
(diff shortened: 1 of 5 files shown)


Platforms/QemuQ35Pkg/PlatformBuild.py
```diff
@@ -200,7 +200,7 @@ def GetPlatformDscAndConfig(self) -> tuple:
 
         The tuple should be (<workspace relative path to dsc file>, <input dictionary of dsc key value pairs>)
         '''
-        return ("QemuQ35Pkg/QemuQ35Pkg.dsc", {})
+        return ("QemuQ35Pkg/QemuQ35PkgX64.dsc", {})
```
makubacki: Should this also return QemuQ35PkgIA32X64.dsc?

makubacki: OvmfPkg does [this](https://github.com/tianocore/edk2/blob/1c74842bd07a48070d1cf3458f0b6b377e3ec8b9/OvmfPkg/PlatformCI/PlatformBuildLib.py#L106-L113):

```python
    def GetPlatformDscAndConfig(self) -> tuple:
        ''' If a platform desires to provide its DSC then Policy 4 will evaluate if
        any of the changes will be built in the dsc.

        The tuple should be (<workspace relative path to dsc file>, <input dictionary of dsc key value pairs>)
        '''
        dsc = CommonPlatform.GetDscName(",".join(self.ActualArchitectures))
        return (f"OvmfPkg/{dsc}", {})
```

kuqin12: Good call. Thanks. This is updated.

PR #1286, opened 2025-12-06
TF-A Patch for Locality Management

Raymond-MS:
## Description

Added support for logical secure partitions in TF-A. Added a logical secure-partition to handle locality management. Initializes the locality states upon start-up. Added code to handle DRTM locality management when measuring to the TPM. Added support for sending DIRECT_REQ2 messages from TF-A. Added support for closing localities 2 and 3 as an SMC to TF-A.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built QEMU SBSA with TPM enabled. Ran to shell successfully. Verified LSP is launched and locality states are updated in the TPM service.

## Integration Instructions

N/A


Platforms/QemuSbsaPkg/tfa_patches/0003-Added-support-for-logical-secure-partitions-in-TF-A.patch
```diff
@@ -0,0 +1,469 @@
...
++				 struct ffa_value *retval);
++
+ uintptr_t plat_spmd_logical_sp_smc_handler(unsigned int smc_fid,
+ 		u_register_t x1,
+ 		u_register_t x2,
+diff --git a/include/services/qemu_sbsa_spmd_logical_sp.h b/include/services/qemu_sbsa_spmd_logical_sp.h
+new file mode 100644
+index 000000000..e832f1be3
+--- /dev/null
++++ b/include/services/qemu_sbsa_spmd_logical_sp.h
```
cfernald: Should this be in the plat/qemu?

cfernald (approved):

Platforms/QemuSbsaPkg/tfa_patches/0003-Added-support-for-logical-secure-partitions-in-TF-A.patch
Raymond-MS: Honestly, I am not sure. I placed this where all of the other logical SP headers are located. It should be part of the include folder, right? If you would like, I can add a qemu section under plat in the include directory and put it there? However, the plan is for this to get upstreamed at some point and this solution is only temporary. We only need this patch because the TPM service will no longer work without the LSP. I think this should be fine to leave as is for now.

Platforms/QemuSbsaPkg/tfa_patches/0003-Added-support-for-logical-secure-partitions-in-TF-A.patch
```diff
@@ -0,0 +1,469 @@
...
+@@ -0,0 +1,12 @@
++/*
++ * Copyright (c) 2023, ARM Limited and Contributors. All rights reserved.
```
kuqin12: Is this something we reused from them?

Platforms/QemuSbsaPkg/tfa_patches/0003-Added-support-for-logical-secure-partitions-in-TF-A.patch
kuqin12: It is fine to have the implementation in the QEMU folder though. Then we will be playing the game of "weak" link thing.

Platforms/QemuSbsaPkg/tfa_patches/0003-Added-support-for-logical-secure-partitions-in-TF-A.patch
Raymond-MS: Yes. They had an "example" logical secure partition that this was based on. I made modifications to implement the features we needed within the LSP but the skeleton is based from ARM.

PR #1318, opened 2026-02-14
Fix MmControlPei does not send MMI to all cores

kuqin12:
## Description

Because there was no negotiation, the SMM controller from QEMU will not broadcast the SMI to all cores, causing only the BSP getting interrupted into MMI.

This change fixed the issue by invoking the negotiation routine and program the register to enable broadcasting.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on Q35 and ensured that all cores are now seen in the MMI.

## Integration Instructions

N/A

### DIFF
--- a/.github/workflows/codeql-platform.yml
+++ b/.github/workflows/codeql-platform.yml
@@ -320,7 +320,7 @@ jobs:
       run: stuart_setup -c ${{ matrix.build_file }} -t DEBUG TOOL_CHAIN_TAG=${{ matrix.tool_chain_tag }}
 
     - name: Upload Setup Log As An Artifact
-      uses: actions/upload-artifact@v6
+      uses: actions/upload-artifact@v7
       if: (success() || failure()) && steps.get_platform_info.outputs.setup_supported == 'true'
       with:
         name: ${{ steps.get_platform_info.outputs.pkg_name }}-Setup-Log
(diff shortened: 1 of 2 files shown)


Platforms/QemuQ35Pkg/SmmControl2Dxe/MmControlPei.c
```diff
@@ -244,6 +244,11 @@ MmControlPeiEntryPoint (
     goto FatalError;
   }
 
+  //
+  // QEMU can inject SMIs in different ways, negotiate our preferences.
+  //
+  NegotiateSmiFeatures ();
```
makubacki: Do you see any need to assert on the return value? I'd assume it should always be `TRUE` and not all cases in `NegotiateSmiFeatures()` goto `FatalError`.

```c
/**
  Negotiate SMI features with QEMU.

  @retval FALSE  It is not an error if SMI feature negotiation is not supported
                 by QEMU. It just means the data cannot be used.

  @retval TRUE   SMI feature negotiation is supported, and it has completed
                 successfully as well (failure to negotiate is a fatal error
                 and the function never returns in that case).
**/
BOOLEAN
NegotiateSmiFeatures (
  VOID
  );
```

For example (would move variable declaration to the beginning of the function):

```suggestion
  BOOLEAN NegotiationSuccessful = NegotiateSmiFeatures ();
  ASSERT (NegotiationSuccessful);
```

makubacki (approved):

PR #1423, opened 2026-05-28
TPM Updates for Q35 & SBSA Including TpmShellApp

Raymond-MS:
## Description

Updating platform level to re-enable TPM on Q35. Add TpmShellApp support for both Q35 and SBSA. Updated the platform to include mapping the CRB region for Q35. Updated the QemuRunner for Q35 to include automatically setting up and running the swtpm. Updated the OemPkg/DeviceBootManagerLib to set up the physical presence variables which are used in the TpmShellApp when attempting to activate PCR banks. Added documentation regarding the TPM on Q35 and SBSA as well as a document about the TpmShellApp and its use on both platforms. Updated the ACPI NV Reclaim memory to support the TPM on Q35. Added a CI workflow to build and run Q35 with the TPM enabled.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Built both QEMU Q35 and SBSA with TPM enabled (TPM2_ENABLE -> Q35 & TPM_ENABLE -> SBSA). Verified TPM functionality. Verified TpmShellApp functionality.

## Integration Instructions

N/A


.github/actions/configure-git/action.yml
```diff
@@ -0,0 +1,20 @@
+# A simple composite action to configure git for Patina QEMU workflows.
```
makubacki:
```suggestion
# A simple composite action to configure git for Mu Tiano Platform QEMU workflows.
```

.github/workflows/q35-tpm.yml
makubacki, line 1: This is a lot of additional CI overhead to test with a feature (TPM) that should be tested by default for modern platform use cases. Can TPM just be added to the normal Q35 build and boot?

.github/workflows/q35-tpm.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Configure Git
+        uses: ./.github/actions/configure-git
+
+      - name: Install dependencies
+        run: pip install -r pip-requirements.txt
+
+      - name: Install swtpm
+        run: |
+          apt-get update
+          apt-get install -y swtpm swtpm-tools
```
makubacki: `swtpm` and `swtpm-tools` need to be cached.

docs/TPM/TpmQemuQ35.md
```diff
@@ -0,0 +1,415 @@
...
+The firmware communicates directly with the TPM device via MMIO, and QEMU forwards the
+I/O to swtpm over a Unix socket.
+
+The base address comes from the SecurityPkg package declaration default
+(`PcdTpmBaseAddress = 0xFED40000`). The Q35 DSC does not override it explicitly. The
+`Tcg2ConfigPei` driver detects the TPM at this address during PEI.
+
+## Architecture Overview
+
+```text
```
makubacki: This is very brittle in terms of becoming stale because it reflects minute implementation details. I recommend not doing this and just encouraging people to generate up-to-date visualizations with AI if they need it. If it stays, it really should be a mermaid diagram.

docs/TPM/TpmQemuQ35.md
```diff
@@ -0,0 +1,415 @@
...
+could coexist. The router accepts only the instance whose `ProviderGuid` matches
+`PcdTpmInstanceGuid`.
+
+### Other DXE Drivers: TCG2 Protocol (`Tpm2DeviceLibTcg2`)
+
+Most DXE drivers that need TPM access use `Tpm2DeviceLibTcg2`, which goes through the
+TCG2 Protocol rather than direct MMIO. Only Tcg2Dxe itself uses `Tpm2DeviceLibRouter`
+with direct MMIO.
+
+### CRB Register Layout
```
makubacki, line 192: This document is long and redundant with existing public information. I suggest not to repeat information verbatim available elsewhere.

docs/TPM/TpmQemuQ35.md
```diff
@@ -0,0 +1,415 @@
...
+| `0xF00` | `Vid` | Vendor ID |
+| `0xF04` | `Did` | Device ID |
+| `0xF08` | `Rid` | Revision ID |
+
+The TIS flow uses `BurstCount` from the STS register to pace reads and writes through
+the `DataFifo` register, one burst at a time.
+
+## Hash Library Architecture
+
+Tcg2Dxe and Tcg2Pei both use `HashLibBaseCryptoRouter` with all five hash instance
```
makubacki: This is information the reader really should find via inspection of the DSC file.

docs/TPM/TpmQemuSbsa.md
```diff
@@ -0,0 +1,667 @@
...
+
+This document describes the TPM 2.0 architecture for the QEMU SBSA platform. SBSA uses
+a dual-CRB design with an FF-A (Firmware Framework for Arm A-Profile) mediated communication
+path between the normal world and the secure world.
+
+Please note that TPM for SBSA is not supported by default and modifications to your QEMU are
+required to get TPM support. These changes are planned to be upstreamed in the future.
+
+For more details please message either Kun Qin <Kun.Qin@microsoft.com> or Raymond Diaz
+<raymond.diaz@microsoft.com> for more information.
```
makubacki: Please do not put names directly in the files. If they want more info, they should create an issue or discussion in the repo and you two reply.

.github/actions/configure-git/action.yml
Raymond-MS: Removed.

.github/workflows/q35-tpm.yml
Raymond-MS: Updated to include TPM support in the base CI workflow, removed the q32-tpm.yml workflow entirely along with the action.yml's it required.

.github/workflows/q35-tpm.yml
Raymond-MS: I think we can leave this as is. The plan should be to include installing this via the Docker Container. Once that is done, we can remove it here.

kuqin12 (approved):

Platforms/QemuQ35Pkg/QemuQ35PkgCommon.dsc.inc
```diff
@@ -872,9 +870,11 @@
   !endif
 
 [PcdsDynamicHii]
-!if $(TPM_ENABLE) == TRUE && $(TPM_CONFIG_ENABLE) == TRUE
+!if $(TPM_ENABLE) == TRUE
+  gEfiSecurityPkgTokenSpaceGuid.PcdTpm2AcpiTableRev|L"TCG2_VERSION"|gTcg2ConfigFormSetGuid|0x8|4|NV,BS
+!endif
+!if $(TPM_CONFIG_ENABLE) == TRUE
```
kuqin12, line 876: I do not think this flag will take effect, right? I vaguely remember it was only used in a PEIM?

Platforms/QemuQ35Pkg/QemuQ35PkgCommon.dsc.inc
```diff
@@ -388,9 +387,8 @@
   TlsLib                     |CryptoPkg/Library/BaseCryptLibOnProtocolPpi/PeiCryptLib.inf
   SmmRelocationLib           |QemuQ35Pkg/Library/SmmRelocationLib/SmmRelocationLib.inf
 !if $(TPM_ENABLE) == TRUE
-  Tpm12DeviceLib             |SecurityPkg/Library/Tpm12DeviceLibDTpm/Tpm12DeviceLibDTpm.inf
   Tpm2DeviceLib              |SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpm.inf
-  Tcg2PreUefiEventLogLib     |QemuPkg/Library/QemuPreUefiEventLogLibNull/QemuPreUefiEventLogLibNull.inf
```
kuqin12, line 392: What is in this old library instance? If no one uses it, shall we remove it?

Platforms/QemuQ35Pkg/QemuQ35PkgCommon.dsc.inc
```diff
@@ -244,9 +244,8 @@
   Tcg2PhysicalPresenceLib |QemuPkg/Library/Tcg2PhysicalPresenceLibNull/DxeTcg2PhysicalPresenceLib.inf
   Tpm2HelpLib             |SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
 !if $(TPM_ENABLE) == TRUE
-  Tcg2PhysicalPresenceLib |QemuPkg/Library/Tcg2PhysicalPresenceLibQemu/DxeTcg2PhysicalPresenceLib.inf
```
kuqin12, line 247: Ditto

docs/TPM/TpmQemuSbsa.md
```diff
@@ -0,0 +1,667 @@
+# TPM on QEMU SBSA
```
kuqin12, line 1: In the spirit of moving away from SBSA, shall we just call it ARM64 virtual platform? With a light touch of SBSA? Otherwise, it will be a major rewording in the next few weeks.

docs/TPM/TpmQemuSbsa.md
```diff
@@ -0,0 +1,667 @@
...
+When `TPM2_ENABLE=TRUE`, the build additionally passes `-DTPM2_ENABLE` to the C compiler
+via build options, allowing C code to use `#ifdef TPM2_ENABLE` guards.
+
+## Platform Memory Layout
+
+The SBSA platform defines two distinct TPM memory regions:
+
+| Region | Address | Size | Visibility |
+| -------- | --------- | ------ | ------------ |
+| Internal CRB | `0x100_00200000` | 0x10 pages | Normal world + Secure world |
```
kuqin12, line 64: I noticed this when I was preparing the change on the Patina-qemu side, these constants have been referenced many times. Can we reword the other instances in this doc to be "internal crb" and "external crb"?

Platforms/QemuQ35Pkg/QemuQ35PkgCommon.dsc.inc
Raymond-MS: DxeTcg2PhysicalPresenceLib is not used in our reference platform. We use DxeTcg2PhysicalPresenceMinimumLib with stripped off functionality including UI and prompting. This library only accepts a subset of commands. As far as I know DxeTcg2PhysicalPresenceLib is used by other platforms.

Platforms/QemuQ35Pkg/QemuQ35PkgCommon.dsc.inc
Raymond-MS: I am not sure what is in this old library instance but if I had to guess since it is a NULL instance that it isn't much, if anything. Not sure if swapping to use the Tcg2 NULL instance would have made much of a difference here but at least we are keeping SBSA and Q35 in sync this way. This may have existed before the SecurityPkg version, I would need to do some more research to be able to give a more informed response.

Platforms/QemuQ35Pkg/QemuQ35PkgCommon.dsc.inc
Raymond-MS: TPM_CONFIG_ENABLE looks to be related to the Tcg2Config DXE driver and the ACPI table. The ACPI table uses PcdTcgPhysicalPresenceInterfaceVer for the SSDT table (at least as far as I could tell). I skimmed Tcg2Config and was not entirely sure what it was doing. As far as I know we never enable TPM_CONFIG_ENABLE. The change here was meant to not guard both lines by both enables, again keeping parity with SBSA.

docs/TPM/TpmQemuSbsa.md
Raymond-MS: Removed.

docs/TPM/TpmQemuSbsa.md
Raymond-MS: I don't think we should change this file. This document is SBSA specific. Once SBSA gets dropped this document could also be dropped. I can generate a new document for VIRT and its TPM when that gets integrated.

docs/TPM/TpmQemuSbsa.md
Raymond-MS: I think this is fine as is as this document is SBSA specific and for SBSA these are our memory regions. Once VIRT is introduced this file will be dropped alongside SBSA and a new document can be generated for VIRT with its TPM information.

makubacki (approved):
I left some comments on the docs, and we had an offline discussion about some of them. I don't have a strong opinion about what you end up doing there if you're committed to monitoring and updating the content over time.

.github/workflows/q35-tpm.yml
makubacki: Since the q35-tpm.yml workflow is also removed, that sounds fine.

docs/TPM/TpmQemuQ35.md
Raymond-MS: Updated to point to the spec instead.

docs/TPM/TpmQemuQ35.md
Raymond-MS: Cleaned up this section.

docs/TPM/TpmQemuQ35.md
Raymond-MS: I would like to keep it as it is very informational. I have tried multiple times to generate a mermaid diagram and every time it has not looked anywhere near as good as this version. Apologies.

makubacki: No problem. Thanks for looking into it.

makubacki (approved):

PR #1429, opened 2026-06-04
QemuRunner TPM Updates for Q35 and SBSA

Raymond-MS:
## Description

Updated QemuRunner.py to no longer point to the /tmp/mytpm1 directory from TPM_DEV, instead it now points to our build output folder. This prevents an issue where creation of the directory is a manual step and running could fail if a user does not do the mkdir step.

Renamed TPM_DEV to SWTPM_ENABLE. SWTPM_ENABLE is TRUE by default. This enable allows for another user, if they so wish, to use a different TPM in place of the SWTPM by setting SWTPM_ENABLE to FALSE.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and ran both Q35 and SBSA with TPM enabled. Verified functionality and swtpm created at the build output directory.

## Integration Instructions

N/A

### DIFF
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -48,22 +48,15 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
 
 
     @staticmethod
-    def RunThread(env):
-        """Runs TPM in a separate thread"""
-        sw_tpm_enable = env.GetValue("TPM_DEV", "FALSE")
-        if str(sw_tpm_enable).upper() == "FALSE":
-            logging.critical("SWTPM Disabled")
-            return
-
-        tpm_dir = env.GetValue("BUILD_OUTPUT_BASE")
-        tpm_sock = os.path.join(tpm_dir, "swtpm-sock")
+    def RunSwTpmThread(tpm_dir, tpm_sock):
+        """Runs SWTPM in a separate thread"""
         tpm_cmd = "swtpm"
         tpm_args = f"socket --tpmstate dir={tpm_dir} --ctrl type=unixio,path={tpm_sock} --tpm2 --log level=1"
 
         # Start the TPM emulator in a separate thread
         ret = utility_functions.RunCmd(tpm_cmd, tpm_args)
         if ret != 0:
-            logging.critical("Failed to start TPM emulator.")
+            logging.critical("Failed to start SWTPM emulator.")
             return
 
     @staticmethod
@@ -219,18 +212,18 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
         args += f" -smbios type=1,manufacturer=Palindrome,product=\"QEMU Q35\",family=QEMU,version=\"{'.'.join(qemu_version)}\",serial=42-42-42-42,uuid=9de555c0-05d7-4aa1-84ab-bb511e3a8bef"
         args += f" -smbios type=3,manufacturer=Palindrome,serial=40-41-42-43{boot_selection}"
 
-        tpm_dev = env.GetValue("TPM_DEV", "FALSE")
-        thread = None
-        if str(tpm_dev).upper() == "TRUE":
+        sw_tpm_thread = None
+        sw_tpm_enable = env.GetValue("SWTPM_ENABLE", "FALSE")
+        if str(sw_tpm_enable).upper() == "TRUE":
             tpm_dir = env.GetValue("BUILD_OUTPUT_BASE")
             tpm_sock = os.path.join(tpm_dir, "swtpm-sock")
             args += f" -chardev socket,id=chrtpm,path={tpm_sock}"
             args += " -tpmdev emulator,id=tpm0,chardev=chrtpm"
             args += " -device tpm-tis,tpmdev=tpm0"
 
-            logging.critical("Starting TPM emulator in a different thread.")
-            thread = threading.Thread(target=QemuRunner.RunThread, args=(env,))
-            thread.start()
+            logging.info("Starting TPM emulator in a different thread.")
+            sw_tpm_thread = threading.Thread(target=QemuRunner.RunSwTpmThread, args=(tpm_dir, tpm_sock))
+            sw_tpm_thread.start()
 
         if (env.GetValue("QEMU_HEADLESS").upper() == "TRUE"):
             args += " -display none"  # no graphics
@@ -287,7 +280,7 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
             # Linux version of QEMU will mess with the print if its run failed, let's just restore it anyway
             utility_functions.RunCmd('stty', 'sane', capture=False)
 
-        if thread is not None:
-            logging.critical("Terminate TPM emulator by using Crtl + C now!")
-            thread.join()
+        if sw_tpm_thread is not None:
+            sw_tpm_thread.join()
+
         return ret
--- a/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
@@ -50,13 +50,15 @@ class QemuRunner(uefi_helper_plugin.IUefiHelperPlugin):
     @staticmethod
     def RunThread(env):
         """Runs TPM in a separate thread"""
-        tpm_path = env.GetValue("TPM_DEV")
-        if tpm_path is None:
-            logging.critical("TPM Path Invalid")
+        sw_tpm_enable = env.GetValue("TPM_DEV", "FALSE")
+        if str(sw_tpm_enable).upper() == "FALSE":
+            logging.critical("SWTPM Disabled")
             return
 
+        tpm_dir = env.GetValue("BUILD_OUTPUT_BASE")
+        tpm_sock = os.path.join(tpm_dir, "swtpm-sock")
         tpm_cmd = "swtpm"
-        tpm_args = f"socket --tpmstate dir={'/'.join(tpm_path.rsplit('/', 1)[:-1])} --ctrl type=unixio,path={tpm_path} --tpm2 --log level=20"
+        tpm_args = f"socket --tpmstate dir={tpm_dir} --ctrl type=unixio,path={tpm_sock} --tpm2 --log level=1"
 
         # Start the TPM emulator in a separate thread
         ret = utility_functions.RunCmd(tpm_cmd, tpm_args)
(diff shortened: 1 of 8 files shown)


kuqin12:
You might want to add the updated TPM_DEV usage in the PR description

kuqin12 (approved):

cfernald (approved):

docs/TPM/TpmQemuQ35.md
makubacki, line 1: @raymond-ms, can you please wait for me to finish a review?

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -217,13 +219,19 @@ def Runner(env):
         args += f" -smbios type=1,manufacturer=Palindrome,product=\"QEMU Q35\",family=QEMU,version=\"{'.'.join(qemu_version)}\",serial=42-42-42-42,uuid=9de555c0-05d7-4aa1-84ab-bb511e3a8bef"
         args += f" -smbios type=3,manufacturer=Palindrome,serial=40-41-42-43{boot_selection}"
 
-        # TPM in Linux
-        tpm_dev = env.GetValue("TPM_DEV")
-        if tpm_dev is not None:
-            args += f" -chardev socket,id=chrtpm,path={tpm_dev}"
+        tpm_dev = env.GetValue("TPM_DEV", "FALSE")
```
spbrogan: not clear why we have two spots.  This also now propagates the default value of "FALSE" to two places.

Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -46,14 +46,16 @@ def QueryQemuVersion(exec):
 
     @staticmethod
     def RunThread(env):
-        ''' Runs TPM in a separate thread '''
-        tpm_path = env.GetValue("TPM_DEV")
-        if tpm_path is None:
-            logging.critical("TPM Path Invalid")
+        """Runs TPM in a separate thread"""
```
spbrogan: why does this ever get called if TPM_DEV == false

Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -46,14 +46,16 @@ def QueryQemuVersion(exec):
 
     @staticmethod
     def RunThread(env):
-        ''' Runs TPM in a separate thread '''
-        tpm_path = env.GetValue("TPM_DEV")
-        if tpm_path is None:
-            logging.critical("TPM Path Invalid")
+        """Runs TPM in a separate thread"""
+        sw_tpm_enable = env.GetValue("TPM_DEV", "FALSE")
```
spbrogan: each user settable variable needs to be defined in project documentation

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -217,13 +219,19 @@ def Runner(env):
         args += f" -smbios type=1,manufacturer=Palindrome,product=\"QEMU Q35\",family=QEMU,version=\"{'.'.join(qemu_version)}\",serial=42-42-42-42,uuid=9de555c0-05d7-4aa1-84ab-bb511e3a8bef"
         args += f" -smbios type=3,manufacturer=Palindrome,serial=40-41-42-43{boot_selection}"
```
makubacki, line 214: The transition from a socket path to a Boolean is strange. If this remains, consider changing the name `TPM_DEV` so this just breaks the old convention and a new environment variable picks up a Boolean type.

docs/TPM/TpmQemuQ35.md
```diff
@@ -29,13 +29,13 @@ See [swtpm Setup](#swtpm-setup) for the full setup commands.
 
 ## Build Configuration
 
-The TPM is disabled by default. To enable it, pass the build define and provide the swtpm
-socket path or provide it in a BuildConfig.conf file placed at the root level of the repo:
+The TPM is disabled by default. To enable it, set `BLD_*_TPM_ENABLE=TRUE` and `TPM_DEV=TRUE`
```
makubacki: Consider a different name than `TPM_DEV` if the continues to be a Boolean as changed in this PR. That is not very descriptive/precise for what is happening. For example, `START_SWTPM=TRUE` is more obvious what is being done.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -217,13 +219,19 @@ def Runner(env):
...
+        thread = None
+        if str(tpm_dev).upper() == "TRUE":
+            tpm_dir = env.GetValue("BUILD_OUTPUT_BASE")
+            tpm_sock = os.path.join(tpm_dir, "swtpm-sock")
+            args += f" -chardev socket,id=chrtpm,path={tpm_sock}"
             args += " -tpmdev emulator,id=tpm0,chardev=chrtpm"
             args += " -device tpm-tis,tpmdev=tpm0"
 
+            logging.critical("Starting TPM emulator in a different thread.")
+            thread = threading.Thread(target=QemuRunner.RunThread, args=(env,))
```
makubacki: My understanding is that if swtpm fails to start a critical error message is printed, but QEMU args still try to connect to swtpm as a client. If that is the case, not attempting the QEMU launch (so the error message is more obvious without QEMU getting in the way) would be clearer to a user.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -217,13 +219,19 @@ def Runner(env):
...
+        tpm_dev = env.GetValue("TPM_DEV", "FALSE")
+        thread = None
+        if str(tpm_dev).upper() == "TRUE":
+            tpm_dir = env.GetValue("BUILD_OUTPUT_BASE")
+            tpm_sock = os.path.join(tpm_dir, "swtpm-sock")
+            args += f" -chardev socket,id=chrtpm,path={tpm_sock}"
             args += " -tpmdev emulator,id=tpm0,chardev=chrtpm"
             args += " -device tpm-tis,tpmdev=tpm0"
 
+            logging.critical("Starting TPM emulator in a different thread.")
```
makubacki: This should be an `info` instead of `critical` message.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -217,13 +219,19 @@ def Runner(env):
         args += f" -smbios type=1,manufacturer=Palindrome,product=\"QEMU Q35\",family=QEMU,version=\"{'.'.join(qemu_version)}\",serial=42-42-42-42,uuid=9de555c0-05d7-4aa1-84ab-bb511e3a8bef"
         args += f" -smbios type=3,manufacturer=Palindrome,serial=40-41-42-43{boot_selection}"
 
-        # TPM in Linux
-        tpm_dev = env.GetValue("TPM_DEV")
-        if tpm_dev is not None:
-            args += f" -chardev socket,id=chrtpm,path={tpm_dev}"
+        tpm_dev = env.GetValue("TPM_DEV", "FALSE")
+        thread = None
+        if str(tpm_dev).upper() == "TRUE":
```
makubacki: Are you currently checking anywhere that is `TPM_DEV` is enabled that `BLD_*_TPM_ENABLE` is set? I don't know if it's necessarily an error, but since two separate params are being maintained trying to identify unintentional mistakes and give guidance would probably help in some cases.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
makubacki, line 1: From a usability standpoint, I'd consider having a single "TPM switch" (e.g. `TPM_ENABLE`) that would enable TPM FW code & launch swtpm. Since we want "TPM enabled by default" (to better reflect modern devices), perhaps the PlatformBuild.py script could detect and set that on qualifying hosts (Linux + swtpm detected).

The actual swtpm-sock path could still be a separate variable, with the previous default. For advanced cases, the user could simply override that to change the path. It could remain `TPM_DEV` or something a little more precise if we want that.

---

Then, eligible systems just build and work with TPM by default, ineligible (Windows host) systems do not, by default and advanced users can just override a single path variable if they need to.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: You are correct, there is no reason to check it twice, removed from RunThread.

Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: It doesn't or it shouldn't. There is no need to check it here. Removed.

Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: Updated the variable. It is documented in the TPM files.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: Updated the name to better reflect its use.

docs/TPM/TpmQemuQ35.md
Raymond-MS: Updated the name to better reflect its use.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: Changed.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: No, I had a discussion with Sean about this. In this instance it allows for the TPM to be enabled but for someone to use a different TPM that isn't the SWTPM if they desire. We specifically do not link the two for this reason. Normal use case will be to have both enabled.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: If the socket is not created, QEMU will fail with an error. I think this is the functionality you are looking for already. There should be no instance where the thread fails to create the socket and QEMU successfully continues without connecting to it.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
Raymond-MS: Created: https://github.com/microsoft/mu_tiano_platforms/issues/1433 to address this at a later date. This change was meant to fix an issue regarding the SWTPM and creating the /tmp/mytpm1 directory which shouldn't be a requirement to get the TPM to work.

I like this idea but I don't want to block this PR on it. Thanks!

kuqin12 (approved):

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
makubacki: As long as the error is clear and at the end of the log, that meets the clarity I had in mind.

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
makubacki: To be clear, the two are not mutually exclusive.

- Your statement (which I agree with): "allows for the TPM (_TPM FW code_) to be enabled but for someone to use a different TPM that isn't the SWTPM if they desire".

- My statement: "if swtpm is used TPM FW code is included"

PR #1430, opened 2026-06-05
Consolidate GOP driver usage for QEMU platforms

kuqin12:
## Description

The current Q35 platform has been using a legacy cirrus video driver, whereas the SBSA is using a PCI based Bochs display.

Given that Q35 also supports the Bochs display, it would make better sense for these platforms to share the same video driver.

The Ramfb driver for Q35 is also removed, which was supposed to be another GOP provider somehow.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on both QEMU Q35 and SBSA and verified the video works properly.

## Integration Instructions

N/A


Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
```diff
@@ -227,7 +227,7 @@ def Runner(env):
         if (env.GetValue("QEMU_HEADLESS").upper() == "TRUE"):
             args += " -display none"  # no graphics
         else:
-            args += " -vga cirrus" #std is what the default is
+            args += " -device bochs-display,addr=0x03 -vga none"
```
vineelko, line 233: >args += " -device bochs-display,addr=0x03 -vga none" [](http://example.com/codeflow?start=12&length=52)

Do we need a similar change in `Platforms/QemuSbsaPkg/Plugins/QemuRunner/QemuRunner.py` for consistency?

`bochs-display,addr=0x1f`

vineelko (approved):

Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py
kuqin12: Good call, but not for SBSA. SBSA machine is default to BOCHS and has internally hardcoded addr, which does not allow override. When we switch to arm-virt, it will need to be specified as you noted.

cfernald (approved):

PR #1443, opened 2026-06-17
QemuRunner: Manage swtpm as a subprocess and wait for its socket

makubacki:
## Description

Makes a few changes to improve stability and robustness of swtpm management.

When a TPM is enabled, the runner starts swtpm and then launches QEMU, which connects to swtpm over a Unix socket. This commit addresses two potential issues.

First, QEMU was started immediately after swtpm with no check that the socket was ready. If swtpm had not yet created the socket before QEMU launch, QEMU failed with "Failed to connect to '...swtpm-sock': No such file or directory".

Second, swtpm was launched by `RunCmd()` inside a Python thread. `RunCmd()` blocks until the process exits, but "swtpm socket" is a long-lived daemon that does not exit on its own, so the thread never finished. The code joined that thread at the end of the run, which could block indefinitely. A thread also gives no handle to the underlying process, so there was no way to stop swtpm or guarantee it was cleaned up from the runner.

Changes:

- Start swtpm with `subprocess.Popen()` and keep the handle, instead of running it through `RunCmd()` in a thread. This gives direct control over the process and avoids the blocking thread join.
- After starting swtpm, poll for the control socket to exist (up to 30s) before launching QEMU, and fail early if swtpm exits or the socket never appears.
- Wrap the QEMU run in `try/finally` so swtpm is always terminated, on success, failure, or exception. `terminate()` is attempted first, then `kill()` if it does not exit within 5 seconds.

This is done for both the Q35 and ArmVirt runners.

---

Note: I saw a `terminate` option for `--ctrl` in swtpm documentation (swtpm will self-terminate when QEMU disconnects) but it was not available in my local swtpm installation (v0.6.3) on Ubuntu 22.04, and although we're using the Ubuntu 24.04 image with swtpm 0.7.3, I think it's okay to just rely on `StopSwTpm()` to terminate it with broader compatibility.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Launch QEMU with swtpm on Ubuntu 22.04 and Ubuntu 24.04 hosts
- Verify swtpm is terminated as expected.

## Integration Instructions

- N/A

### DIFF
--- a/Platforms/QemuArmVirtPkg/Plugins/QemuRunner/QemuRunner.py
+++ b/Platforms/QemuArmVirtPkg/Plugins/QemuRunner/QemuRunner.py
@@ -11,7 +11,8 @@ import io
 import os
 import re
 import datetime
-import threading
+import subprocess
+import time
 from pathlib import Path
 from edk2toolext.environment.plugintypes import uefi_helper_plugin
 from edk2toollib import utility_functions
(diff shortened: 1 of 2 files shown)


apop5 (approved):

kuqin12 (approved):

makubacki:
This is primarily to address the high frequency (>50% failures) seen in Mu PR Validation workflows:

```
INFO - qemu-system-x86_64: -chardev socket,id=chrtpm,path=/__w/mu_plus/mu_plus/Build/QemuQ35PkgX64/DEBUG_CLANGPDB/swtpm-sock: Failed to connect to '/__w/mu_plus/mu_plus/Build/QemuQ35PkgX64/DEBUG_CLANGPDB/swtpm-sock': No such file or directory
```

These have been observed to happen for both QemuQ35Pkg and QemuArmVirtPkg. I'm running the changes in this PR with the full Mu PR Validation workflow several times on a fork to confirm that particular failure is handled and then I'll merge the PR.

makubacki:
> I'm running the changes in this PR with the full Mu PR Validation workflow several times on a fork to confirm that particular failure is handled and then I'll merge the PR.

5/5 runs were successful.

PR #1450, opened 2026-06-23
TPM Platform Updates

### LINKED ISSUE #1433: Improve TPM Usability:
From a usability standpoint, I'd consider having a single "TPM switch" (e.g. `TPM_ENABLE`) that would enable TPM FW code & launch swtpm. Since we want "TPM enabled by default" (to better reflect modern devices), perhaps the PlatformBuild.py script could detect and set that on qualifying hosts (Linux + swtpm detected).

The actual swtpm-sock path could still be a separate variable, with the previous default. For advanced cases, the user could simply override that to change the path. It could remain `TPM_DEV` or something a little more precise if we want that.

---

Then, eligible systems just build and work with TPM by default, ineligible (Windows host) systems do not, by default and advanced users can just override a single path variable if they need to.

_Originally posted by @makubacki in https://github.com/microsoft/mu_tiano_platforms/pull/1429#discussion_r3367934822_

Raymond-MS:
## Description

Updated the build-haf-tfa worflow to generate both the TPM enabled and default builds. Updated the platform-ci workflow to include ARM VIRT as a TPM build. Updated Q35 to use TPM2_ENABLE instead of TPM_ENABLE. Updated the documentation to reflect the change to TPM2_ENABLE. Updated the QemuRunner.py scripts to generate a descriptive error when SWTPM is not installed on Linux builds not using the container. Updated the secure_partition_binaries_ext_dep.yaml to point to the latest v1.0.0 release files.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Built both Q35 and ARM VIRT with TPM enabled. Verified functionality on both using the TpmShellApp.

## Integration Instructions

Update your Q35 builds to use TPM2_ENABLE rather than TPM_ENABLE.


Raymond-MS:
With all current changes, this should address https://github.com/microsoft/mu_tiano_platforms/issues/1433. A single TPM2_ENABLE now works for both Q35 and ARM VIRT. SWTPM is enabled by default. It is installed by default in the container and for developers not using the container, a detailed message is displayed if they do not have it installed.

Platforms/QemuArmVirtPkg/PlatformBuild.py
```diff
@@ -840,6 +840,28 @@ def PlatformPreBuild(self):
         return 0
 
     def PlatformPostBuild(self):
+        # Allow developers to point the mssp-rust SP at a locally-built .bin
+        # by setting MSSP_RUST_BIN_OVERRIDE.
+        override = (self.env.GetValue("MSSP_RUST_BIN_OVERRIDE") or "").strip()
```
kuqin12: If the user overrides the path, they would directly set the `MSSP_RUST_BIN_FILE` right? Then we just need to check to see if the user has set this before trying to setting the default value?

Platforms/QemuArmVirtPkg/PlatformBuild.py
```diff
@@ -851,7 +873,13 @@ def PlatformPostBuild(self):
             if not ext_dep_bins:
                 logging.error("HAF_TFA_BINS not set. Cannot patch secure partitions.")
                 return -1
-            op_tfa = Path(ext_dep_bins)
+
+            if tpm2_enable:
+                variant_dir = "TPM"
+            else:
+                variant_dir = "DEF"
```
kuqin12, line 874: nit: I guess trying to enable TPM on `nt` machines are not supported?

kuqin12 (approved):

Platforms/QemuArmVirtPkg/PlatformBuild.py
Raymond-MS: It is not supported no. We disable the SWTPM on nt builds. I don't see why you can't build but you will definitely not be able to run.

Platforms/QemuArmVirtPkg/PlatformBuild.py
Raymond-MS: We don't set a default for MSSP_RUST_BIN_FILE. If there is an override, MSSP_RUST_BIN_FILE points to what the user set MSSP_RUST_BIN_OVERRIDE to otherwise depending on TPM2_ENABLE it will either be the TPM build or the default build.

Platforms/QemuArmVirtPkg/PlatformBuild.py
kuqin12: Good, then we might want to add a check and just fail the build here. :)

makubacki:
@raymond-ms, I've become invested enough in these changes between here and patina-qemu that I'd like to take a quick look through tomorrow if it can wait.

Platforms/QemuArmVirtPkg/PlatformBuild.py
kuqin12: There is no need for 2 different build variables. Just `MSSP_RUST_BIN_FILE` will do the job.

kuqin12:
Also, this is a breaking change because the TPM_ENABLE -> TPM2_ENABLE will require the user to change the flag manually. Please mark the PR accordingly.

Platforms/QemuArmVirtPkg/PlatformBuild.py
makubacki: I haven't looked through the full PR yet so larger context might be missing, but let's try not to put artificial barriers in place. If the only problem is lack of a software TPM (e.g. swtpm), please let the build occur and fail cleanly based on SWTPM_ENABLE.

Platforms/QemuQ35Pkg/QemuQ35PkgX64.dsc
```diff
@@ -41,11 +41,11 @@
   MSFT:*_*_X64_GENFW_FLAGS  = --keepexceptiontable
   GCC:*_*_X64_GENFW_FLAGS   = --keepexceptiontable
 
-!if $(TPM_ENABLE) == TRUE
+!if $(TPM2_ENABLE) == TRUE
   #
   # Enable TPM support
   #
-  *_CLANGPDB_*_CC_FLAGS = -DTPM_ENABLE
+  *_CLANGPDB_*_CC_FLAGS = -DTPM2_ENABLE
```
makubacki: We still have VS & GCC builds running and compiler flags for those tool chains used in this file. Is there a reason not to set flag for those too?

Platforms/Docs/Common/Features/feature_tpm.md
makubacki: I think it would be helpful to call out here that the container has swtpm installed by default and that it is recommended to use the container, so all software versioning (including swtpm) matches the versions validated in CI.

makubacki (approved):

Platforms/QemuArmVirtPkg/PlatformBuild.py
kuqin12: Maybe I am out of sync with the current state. But the confusion is with `TPM2_ENABLE = TRUE` + `SWTPM_ENABLE = FALSE` + `nt`. The build will proceed and not failing on the absence of swtpm but assert at runtime when UEFI is trying to talk to TPM device.

I agree that blocking the build seems odd because it would prevent the alternative software TPM to be used. But I take this as a tradeoff before we can knowingly support other tpm backends.

Platforms/QemuArmVirtPkg/PlatformBuild.py
Raymond-MS: Removed.

Platforms/QemuArmVirtPkg/PlatformBuild.py
makubacki: I think we're saying the same thing. What I'm saying is:

- `TPM2_ENABLE` should always build on `nt` (has nothing to do with using software TPM at runtime)
- `SWTPM_ENABLE` should always fail on `nt` and clearly say a TPM emulator is not support on `nt` because it is hardcoded to launch `swtpm` today:
  https://github.com/microsoft/mu_tiano_platforms/blob/2f0a2be4cedc5f952b6662662fd0005e447be7bb/Platforms/QemuQ35Pkg/Plugins/QemuRunner/QemuRunner.py#L51-L66

Raymond-MS: I think a good middle ground is to not fail the build but fail in QemuRunner if `TPM2_ENABLE=TRUE`. We know that for `nt` builds `SWTPM_ENABLE=FALSE` so if a user sets `TPM2_ENABLE=TRUE` they may successfully build but it will fail during run. Adding a check here with a log makes it so it fails fast on `nt` builds. My thought process is:

`TPM2_ENABLE=TRUE` + `SWTPM_ENABLE=TRUE` -> GOOD
`TPM2_ENABLE=FALSE` + `SWTPM_ENABLE=TRUE` -> GOOD
`TPM2_ENABLE=FALSE` + `SWTPM_ENABLE=FALSE` -> GOOD
`TPM2_ENABLE=TRUE` + `SWTPM_ENABLE=FALSE` -> BAD

`SWTPM_ENABLE=TRUE` by default except for `nt` builds. If a user manually sets `SWTPM_ENABLE=FALSE` they are actively introducing an issue unless they replace the backend. In this instance, we do not need to check because it would be our assumption that a user knows what they are doing setting `SWTPM_ENABLE=FALSE` manually. Once we have a TPM solution for `nt` builds we can remove this code.

kuqin12: I was under the impression that the boot will fail because of the mismatch between the build and runtime flags. But Raymond and I looked at this offline and confirmed that the boot fail only applies to ARM virt.

With that, I think we do not need to drill this topic too hard. But the real underlying is once again the build needs to match the runtime setup, just like the core count. I think we should untangle that as the mid-term resolution.

The solution after that will be to onboard Windows environment to support software TPM. Then we will drop this TPM enable flag completely.

For now, this is good as is.

Platforms/QemuQ35Pkg/QemuQ35PkgX64.dsc
Raymond-MS: Good catch updated to include both GCC and MSFT which should handle all instances. This was an oversight on my end when I ported this from patina-qemu.

Platforms/Docs/Common/Features/feature_tpm.md
Raymond-MS: Added.

Platforms/QemuArmVirtPkg/PlatformBuild.py
makubacki: I spoke to Raymond offline too and we agreed on the main points about build vs runner and mixing/matching binaries and boot across environments. As long as the mainstream use cases are supported and limitations documented, I think it's fine.

PR #1455, opened 2026-06-26
Update mu_plus and mu_basecore

os-d:
## Description

Update mu_plus so that QEMU PR validations will succeed with a new basecore change incoming. Update basecore while we are here.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.


makubacki:
You're pulling in https://github.com/microsoft/mu_tiano_platforms/pull/1454 (which is fine) but could you also pull in https://github.com/microsoft/mu_tiano_platforms/pull/1453? We can just get the submodule updates out of the way here in a single PR.

os-d:
> You're pulling in #1454 (which is fine) but could you also pull in #1453? We can just get the submodule updates out of the way here in a single PR.

I was just updating mu_plus in my current PR because I believe this set of steps needs to happen:
- Update mu_plus in the two platform repos -> get QEMU validation working with the new basecore PR that is going in
- mu_plus override validation is going to start failing because we went that route in PR feedback, go back and update the override tag for the new basecore PR
- Update mu_plus and basecore in platform repos to get to a steady state

os-d:
Went ahead and updated now, no harm other than CI time.

makubacki (approved):

apop5 (approved):

